PR #8, opened 2021-08-19
Improvements

jordeu:
- Update to latest Tower API
- Use GNU style CLI options
- Refactor to use annotations to add subcommands
- Add top level options
- Add x-ray option to show request/response data and headers at stderr


pditommaso (approved):
Ok, I went in auto-approval mode

PR #24, opened 2021-09-29
Task 18 / Health Check

grananda:


src/main/java/io/seqera/tower/cli/responses/HealthCheckResponse.java
```diff
@@ -0,0 +1,34 @@
...
+        this.credentialsCheck = credentialsCheck;
+    }
+
+    @Override
+    public void toString(PrintWriter out) {
+        String ok = ansi("@|fg(green) OK|@");
+        String fail = ansi("@|fg(red) FAILED|@");
+        String undefined = ansi("@|fg(yellow) UNDEFINED|@");
+
+        out.println(ansi("    @|bold,fg(yellow) System health status|@"));
```
jordeu, line 25: Prefix it with a `%n` to leave the first line empty (like other responses)

src/main/java/io/seqera/tower/cli/commands/HealthCheckCmd.java
```diff
@@ -0,0 +1,49 @@
...
+        name = "health",
+        description = "Checks system health status"
+)
+public class HealthCheckCmd extends AbstractRootCmd {
+    @Override
+    protected Response exec() throws ApiException, IOException {
+        int connectionCheck = 1;
+        int versionCheck = -1;
+        int credentialsCheck = -1;
+        int requiredApiVersion = 160;
```
jordeu, line 21: Add the VersionProvider to get the minimum required API version. It can be accessed like this: `app().spec.versionProvider()`. 
http://github.com/seqeralabs/tower-cli/blob/8562434d079f34e7fe756301b06d57e7c734c526/src/main/java/io/seqera/tower/cli/utils/VersionProvider.java#L7-L15

jordeu: You'll need to add a new `getAPIVersion` method at the `VersionProvider`.

src/main/java/io/seqera/tower/cli/commands/HealthCheckCmd.java
```diff
@@ -0,0 +1,49 @@
...
+
+        try {
+            ServiceInfoResponse infoResponse = api().info();
+
+            if (infoResponse.getServiceInfo() != null && infoResponse.getServiceInfo().getApiVersion() != null) {
+                versionCheck = 0;
+                if (Integer.parseInt(infoResponse.getServiceInfo().getApiVersion().replace(".", "")) >= requiredApiVersion) {
```
jordeu, line 28: Consider using `ModuleDescriptor.Version` to compare the versions. 
https://docs.oracle.com/javase/9/docs/api/java/lang/module/ModuleDescriptor.Version.html

src/main/java/io/seqera/tower/cli/responses/HealthCheckResponse.java
```diff
@@ -0,0 +1,34 @@
...
+
+        TableList table = new TableList(out, 2);
+        table.setPrefix("    ");
+        table.addRow("Remote API server connection check", connectionCheck == 1 ? ok: connectionCheck == 0 ? fail: undefined);
+        table.addRow("Tower API version check", versionCheck == 1 ? ok: versionCheck == 0 ? fail: undefined);
+        table.addRow("Authentication API credential's token", credentialsCheck == 1 ? ok: credentialsCheck == 0 ? fail: undefined);
+        table.print();
```
jordeu, line 32: Below the table it will be interesting to show one or multiple messages when some of the checks fail. For example:

- Tower API URL https://tower.nf it is not available (did you mean https://tower.nf/api?)
- Tower API version is 1.5.1 while the minimum required version to be fully compatible is 1.6
- Review that your current access token is valid and active.

src/main/java/io/seqera/tower/cli/responses/HealthCheckResponse.java
```diff
@@ -0,0 +1,34 @@
+package io.seqera.tower.cli.responses;
+
+import io.seqera.tower.cli.utils.TableList;
+
+import java.io.PrintWriter;
+
+public class HealthCheckResponse extends Response{
```
jordeu, line 7: To use this command inside a bash script, it's interesting to return an `exitCode` equal to -1 when there is at least one fail (otherwise return a zero).

Currently it is not possible to set the exit code of a response, but we can add a `getExitCode` method to the abstract class `Response` that by default returns a zero and override it here.

jordeu (changes requested):

src/main/java/io/seqera/tower/cli/commands/HealthCheckCmd.java
grananda: Sorry, forgot that one!

jordeu (approved):

PR #26, opened 2021-10-05
Task 25 / Add action command

grananda:


jordeu:
Maybe for now it make sense, because this is what we have at API level:
```
class CreateActionRequest {

    String name
    Action.Source source
    WorkflowLaunchRequest launch

}
```
So even if the github action has configuration:
```
class GithubActionConfig implements Action.ConfigType {
    List<String> events

    @Override
    String getDiscriminator() {
        return Action.Source.github.toString()
    }
}
```
we are just harcoding it to only add "push" events and we cannot set the action configuration using the current API.

When proposing to use subcommands like `towr actions create github ...` and `towr actions create tower ...` instead of `towr actions create --source [github/tower] ...` I was thinking more in a future when we will have more types of actions (for example [S3 events](https://docs.aws.amazon.com/AmazonS3/latest/userguide/NotificationHowTo.html) or similar) with different configurations.

Maybe @pditommaso  @mbosio85 can help us decide if it is worth it to use the subcommand approach right now or we don't expect to have more actions types in a near future so we can go on with `towr actions create --source ...`  approach (notice that with this approach in a future if we have more types with different configurations we'll need to break CLI backwards compatibility to add them).

jordeu (changes requested):

PR #77, opened 2021-10-28
Task 28 / Import and export for pipelines and ce

### LINKED ISSUE #28: Allow to import/export pipelines and compute-envs from json
On possible solution could be:

To export:

`towr pipelines view -i <pipeline_id> --export`

This will print the *json* to *stdout*

_NOTE: we can replicate the same option `--export` at `compute-envs`_ 

To import:

`towr pipelines create -n <pipeline_name> --json <path_to_export_json_file>`

grananda:
Closes #28


src/main/java/io/seqera/tower/cli/responses/ComputeEnvs/ComputeEnvExport.java
```diff
@@ -0,0 +1,35 @@
...
+    public String toString() {
+        String configOutput = "";
+
+        try {
+            configOutput = new JSON().getContext(CreateComputeEnvRequest.class).writerWithDefaultPrettyPrinter().writeValueAsString(request);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
+        }
+
+        if (fileName != null) {
```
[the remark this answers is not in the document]
grananda: > Not sure if we need to allow to export into a file on this first iteration.

I believe the export to file is important because it allows you to execute and export and import command set in a non-sequential way by using the file name as a reference. I think it is a matter of how do we wish to use the E/I commands. Nevertheless, the export command does not require a `filename` to be defined.

> I'd add the option to accept `-` as a possible `fileName` value.

Don't quite understand this, sorry.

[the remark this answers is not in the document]
jordeu: Using `-` as a filename to mean `stdin` or `stdout` is a convention that a lot of command line programs use. So if the user sets `-` as filename the expected behaviour is to print to `stdout` (the same that we do if no filename is given)

src/main/java/io/seqera/tower/cli/responses/pipelines/PipelinesExport.java
```diff
@@ -0,0 +1,46 @@
...
+    public String toString() {
+        String configOutput = "";
+
+        try {
+            configOutput = new JSON().getContext(CreatePipelineRequest.class).writerWithDefaultPrettyPrinter().writeValueAsString(createPipelineRequest);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
+        }
+
+        if (fileName != null) {
```
jordeu, line 40: Here the logic has to be the same that now we are using at `compute-envs export`. Accept the dash `-` as a synonymous of `stdout` and give a message that the pipeline has been exported when using a file.

src/main/java/io/seqera/tower/cli/responses/pipelines/PipelinesExport.java
```diff
@@ -0,0 +1,46 @@
...
+            FilesHelper.saveString(fileName, configOutput);
```
jordeu, line 41: The save of the file has to be done at the `ExportCmd` instead of here. The idea is that the responses only do presentation, no business logic. In fact, the `toString` method it is not call when the user wants always `json` output. (when using `tw --json ...`)

src/main/java/io/seqera/tower/cli/responses/ComputeEnvs/ComputeEnvExport.java
```diff
@@ -0,0 +1,48 @@
...
+        String configOutput = "";
+
+        try {
+            configOutput = new JSON().getContext(ComputeConfig.class).writerWithDefaultPrettyPrinter().writeValueAsString(computeConfig);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
+        }
+
+        if (fileName != null && !fileName.equals("-")) {
+            FilesHelper.saveString(fileName, configOutput);
```
jordeu, line 41: The save of the file has to be done at the `ExportCmd` instead of here. The idea is that the responses only do presentation, no business logic. In fact, the `toString` method it is not call when the user wants always `json` output. (when using `tw --json ...`)

jordeu (changes requested):

src/main/java/io/seqera/tower/cli/responses/ComputeEnvs/ComputeEnvExport.java
grananda: To me, a file is a way to present data. Under that criteria, should we move the JSON conversion to the command as well?

jordeu: Yes, you're right. The JSON conversion also needs to be moved to the command as well.

src/main/java/io/seqera/tower/cli/commands/pipelines/ImportCmd.java
```diff
@@ -0,0 +1,58 @@
...
+        CreatePipelineRequest request;
+
+        request = parseJson(FilesHelper.readString(fileName), CreatePipelineRequest.class);
+        request.setName(name);
+
+        if (computeEnv != null) {
+            ComputeEnv ce = findComputeEnvironmentByName(computeEnv, workspaceId());
+            request.getLaunch().setComputeEnvId(ce.getId());
+        }
+
```
jordeu, line 53: When no `computeEnvId` is provided at the `json` neither as a command line option, the API returns a http 403 error and then the CLI shows a cryptic error message 
```
ERROR: Unknown. Check that the provided identifier is correct.
```

May be, here we can check that the CE is provided and fail with a nice error message if not.

jordeu (changes requested):

src/main/java/io/seqera/tower/cli/commands/pipelines/ImportCmd.java
grananda: As far as I can see, a CE will always be available in the import `json` file unless y manually remove it, right? Will this scenario ever occur?

jordeu: For me makes sense to manually remove the CE from the `json` if you plan to import that pipeline in many different workspaces.

jordeu: in fact I've also seen that when you try to import a pipeline with a CE from another workspace you also get the same cryptic error:
```
ERROR: Unknown. Check that the provided identifier is correct.
```
So we need to improve this message checking if the user has access to that CE, or opening an issue at Tower repo if you thing that this has to be improved a API side.

jordeu (approved):
Reviewing this I found a possible bug at API level https://github.com/seqeralabs/nf-tower-cloud/issues/2204 that still makes the cryptic error message appear on some cases.

But, if it's fixed a API side, the error message will be meaningful for the user. So everything seems ok now.

swampie (approved):

PR #90, opened 2021-11-04
Task 64 / Remove workspace option as global

grananda:
Workspace and organization ID and name are now specific as per command and not used as global parameters. This will improve the autogenerated help documentation and unify the API and the CLI when it comes to parameter sync.


src/main/java/io/seqera/tower/cli/Tower.java
```diff
@@ -65,12 +64,6 @@
     @Option(names = {"-t", "--access-token"}, description = "Tower personal access token (TOWER_ACCESS_TOKEN)", defaultValue = "${TOWER_ACCESS_TOKEN}")
```
jordeu, line 64: Not related to this issue, but I think that we need to make this field required.

src/main/java/io/seqera/tower/cli/commands/global/WorkspaceOptions.java
```diff
@@ -0,0 +1,20 @@
...
+ * defined by the Mozilla Public License, v. 2.0.
+ */
+
+package io.seqera.tower.cli.commands.global;
+
+import picocli.CommandLine;
+
+public class WorkspaceOptions {
+
+    @CommandLine.Option(names = {"-w", "--workspace"}, description = "Workspace numeric identifier (TOWER_WORKSPACE_ID)", defaultValue = "${TOWER_WORKSPACE_ID}")
```
jordeu, line 18: When this `mixin` is added the workspace is always required. I think we need to set `required = true` here.

jordeu (changes requested):

src/main/java/io/seqera/tower/cli/commands/global/WorkspaceOptions.java
grananda: If I make it required, it will still read the env if omitted but will throw an error if the env is not set, which is not desirable when working with user's resources.

src/main/java/io/seqera/tower/cli/commands/AbstractApiCmd.java
```diff
@@ -95,25 +95,25 @@ protected void performAdditionalClientConfiguration(ClientConfig clientConfig) {
...
-                if (app().orgAndWorkspaceNames != null) {
-                    loadOrgAndWorkspaceFromNames();
-                }
-            }
-        }
-        return workspaceId;
-    }
-
-    protected Long orgId() throws ApiException {
+//    protected Long workspaceId() throws ApiException {
```
swampie, line 98: any reason why the code is commented? If it is unused and you don't plan to revert this change just delete it.

src/main/java/io/seqera/tower/cli/commands/LaunchesCmd.java
```diff
@@ -49,7 +54,7 @@
     @Option(names = {"-c", "--compute-env"}, description = "Compute environment name (defaults to workspace primary)")
     String computeEnv;
 
-    @Option(names = {"-w", "--work-dir"}, description = "Working directory")
+    @Option(names = {"--work-dir"}, description = "Working directory")
```
swampie, line 57: mmmh: I don't like that the short version of the argument has been removed. Can we use -d because it is not consistent having all the other with the short and long version while this one only has the long option.

src/main/java/io/seqera/tower/cli/commands/computeenvs/platforms/AbstractPlatform.java
```diff
@@ -23,10 +23,12 @@
 
 public abstract class AbstractPlatform<T extends ComputeConfig> implements Platform {
 
-    @Option(names = {"-w", "--work-dir"}, description = "Work directory", required = true)
+    @Option(names = {"--work-dir"}, description = "Work directory", required = true)
```
swampie, line 26: same comment as above

src/main/java/io/seqera/tower/cli/commands/LaunchesCmd.java
grananda: Some params have no short version:

https://github.com/seqeralabs/tower-cli/blob/510a77f39e8a330bf9f2cb353f0e51044d19382f/src/main/java/io/seqera/tower/cli/commands/computeenvs/platforms/AwsBatchManualPlatform.java#L25-L30

In addition `-d` has conflicts with the `description` param, but `-k` is available.

src/main/java/io/seqera/tower/cli/commands/global/WorkspaceOptions.java
jordeu: What do you mean "working with user's resources"?

grananda: A compute environment that is not under a given workspace, for instance.

src/main/java/io/seqera/tower/cli/commands/AbstractApiCmd.java
grananda: Sorry...was there as a preventive measure but it is removed

src/main/java/io/seqera/tower/cli/commands/global/WorkspaceOptions.java
jordeu: Ok, I see. Yes, you're right. Better to keep it optional.

jordeu (commented):
As it is now, when you show the help of a command the `-w` is always the first, even before other command options that are required. I think that the usage will be more clear if we move the workspace mixin as the last option, or at least after the required ones depending on the command.

src/main/java/io/seqera/tower/cli/commands/LaunchesCmd.java
jordeu: I really don't like to do not use `-w` as a shortcut for `--work-dir`. The main reason is because we are trying to use as much as possible the same options that is using `nextflow run` command line. 

I was thinking that if in a future we implement the "profile" feature so that you can switch to permanently use a different workspace, most likely the `--workspace` option won't be used so much. My proposal would be to change the workspace option to use this `"-ws", "--workspace"` or maybe this `"--ws", "--workspace"`. And then keep using `-w` for `workDir`. Opinions?

pditommaso: `-w`, `--work-dir `

and 

`-W`, `--workspace` 

?

jordeu: Yes, nice. I didn't thought about this option. I like it.

pditommaso: tho, I think it's much more needed to being able to specify the workspace, more than the work dir. Would not make sense to only have `--work-dir` and use `-w` for the workspace?

jordeu: Currently you can also specify the workspace using the `TOWER_WORKSPACE_ID` environment variable. And in a future I imagine something similar to AWS CLI that you can define different "profiles" and switch between them so you can define the new active profile. And then you don't need to pass the workspace on all the commands while you are working on the same workspace, just once to change the active workspace. If we've this, then I think that to pass the workspace using command line options won't be used so much. This is the main reason why I think that it's a petty to do not use `-w` for work directory as in `nextflow run`.

jordeu: I imagine a command similar to this to switch context:
```
tw set-context --current --workspace=<...> --token=<...> --url=<...>
```

grananda: > This is the main reason why I think that it's petty to do not use `-w` for the work directory as in `nextflow run`.

Still, we may wish to keep the use of `-w` to define a workspace on the fly despite the profile feature. Also, for backward compatibility.

Personally, I would stick with `--work-dir `(and perhaps `-k`, since it is not compromised) and keep `-w` and `-o `for workspace and organization.

Keep in mind we do not need to have a single word param shortcut for all options, we wouldn't have enough letters.

jordeu:
For me, to use `-k` to refer to the `--work-dir` is the worst option. I prefer to not give any short option to `--work-dir` or use the `-W`/`-w` options before. I think that we can quickly discuss this in the dev meeting just to decide what do we all prefer.

jordeu (approved):

PR #98, opened 2021-11-08
Task 34 / Primary compute env operations

grananda:


swampie (approved):

src/main/java/io/seqera/tower/cli/responses/ComputeEnvs/ComputeEnvsPrimaryGet.java
```diff
@@ -0,0 +1,31 @@
...
+
+import io.seqera.tower.cli.responses.Response;
+import io.seqera.tower.model.ComputeEnv;
+
+public class ComputeEnvsPrimaryGet extends Response {
+
+    public final String workspaceRef;
+    public final ComputeEnv computeEnv;
+
+    public ComputeEnvsPrimaryGet(String workspaceRef, ComputeEnv computeEnv) {
```
jordeu, line 22: When using the option `--json` the whole `computeEnv` object is show. I think that it would be better to here only pass the `computeEnv.id` and `computeEnv.name` so the `json` output is less verbose.

src/main/java/io/seqera/tower/cli/responses/ComputeEnvs/ComputeEnvsPrimarySet.java
```diff
@@ -0,0 +1,31 @@
...
+
+import io.seqera.tower.cli.responses.Response;
+import io.seqera.tower.model.ComputeEnv;
+
+public class ComputeEnvsPrimarySet extends Response {
+
+    public final String workspaceRef;
+    public final ComputeEnv computeEnv;
+
+    public ComputeEnvsPrimarySet(String workspaceRef, ComputeEnv computeEnv) {
```
jordeu, line 22: When using the option `--json` the whole `computeEnv` object is show. I think that it would be better to here only pass the `computeEnv.id` and `computeEnv.name` so the `json` output is less verbose.

jordeu (changes requested):

src/main/java/io/seqera/tower/cli/responses/ComputeEnvs/ComputeEnvsPrimaryGet.java
grananda: Not sure about this, after all, you are requesting to VIEW a specific compute environment, just like we do with the VIEW command when requesting by ID. If a user wishes to just check what the primary CE is, she can just use the LIST command with the new star feature.

I believe this command is far beyond checking what my CE is. Let me know what you think.

swampie: Also you can use a tool like jq to parse and eventually trim the results down

grananda: But, do we want to trim the result? Maybe I wish to VIEW the full primary  CE object to pipe it to another command.

swampie: No absolutely. For me the result of the cli should be complete: than it's up to the user to eventually trim the result down by piping the result into another command like `jq`. I am curious though about the difference between the export and the view command now. I'll do some tests

src/main/java/io/seqera/tower/cli/responses/ComputeEnvs/ComputeEnvsPrimarySet.java
grananda: I have added the following short CE version, let me know if it is to much:

```
{
  "credentials_id" : "G0lOxITHSVeAXhoKOU9uR",
  "name" : "AWS_Batch_eu-west-1",
  "id" : "1voNN3Mve8r49Ek6hFCN5v",
  "platform" : "aws-batch"
}
```

jordeu (approved):

PR #99, opened 2021-11-08
Task 52 / Change json flag

grananda:
JSON option flag was modified by --output in order to allow multiple output parameters in the futute.


src/main/java/io/seqera/tower/cli/Tower.java
```diff
@@ -67,8 +67,8 @@
     @Option(names = {"-u", "--url"}, description = "Tower server API endpoint URL. Defaults to tower.nf (TOWER_API_ENDPOINT)", defaultValue = "${TOWER_API_ENDPOINT:-https://api.tower.nf}", required = true)
     public String url;
 
-    @Option(names = {"-j", "--json"}, description = "Show output as JSON")
-    public boolean json;
+    @Option(names = {"--output"}, description = "Show output in defined format (only JSON is available so far)")
+    public String output;
```
jordeu, line 71: Use an enum here. In that way picocli validates the value.

src/main/java/io/seqera/tower/cli/Tower.java
```diff
@@ -67,8 +67,8 @@
     @Option(names = {"-u", "--url"}, description = "Tower server API endpoint URL. Defaults to tower.nf (TOWER_API_ENDPOINT)", defaultValue = "${TOWER_API_ENDPOINT:-https://api.tower.nf}", required = true)
     public String url;
 
-    @Option(names = {"-j", "--json"}, description = "Show output as JSON")
-    public boolean json;
+    @Option(names = {"--output"}, description = "Show output in defined format (only JSON is available so far)")
```
jordeu, line 70: Improve the description, it is not clear if the user has to pass `--output JSON`  or `--output json`.

jordeu (changes requested):

src/main/java/io/seqera/tower/cli/commands/AbstractApiCmd.java
```diff
@@ -260,7 +261,7 @@ protected String workspaceRef(Long workspaceId) throws ApiException {
     public Integer call() {
         try {
             Response response = exec();
-            if (app().json) {
+            if (app().output != null && Objects.equals(app().output, OutputType.json)) {
```
jordeu, line 264: I think this can be simplify to only `app().output == OutputType.json`

grananda: It was a fancy Intellij suggestion... it pisses me off when the IDE underlines the code in orange.. will revert :)

jordeu (approved):

PR #104, opened 2021-11-10
Task 100 / feature: Create metrics subcommand

grananda:
Metrics were move to an individual subcommand from view


grananda:
Regarding the usage of the wildcard (*) option to designate all the metric types and columns, I have opted to use an `Enum`, where no options will lead to all options (*). The user will have to specify the columns and types parameter in case they choose to request fewer options.

jordeu (changes requested):
To make the table more user friendly we need to:

1. format the numbers using units 
2. default columns order must be: min, q1, median (q2), mean, q3, max
3. collapse all the data of one process into a single row.

I'll add examples of each table on following comments.

jordeu:
**Memory section**

From:

```
     Memory Metrics
    ----------------

     process                        | metric     | mean         | min          | max          | q1           | q2           | q3           
    --------------------------------+------------+--------------+--------------+--------------+--------------+--------------+--------------
     SAMPLESHEET_CHECK              | memVirtual | 3.28499E7    | 3.28499E7    | 3.28499E7    | 3.28499E7    | 3.28499E7    | 3.28499E7    
     SAMPLESHEET_CHECK              | memRaw     | 9662460.0    | 9662460.0    | 9662460.0    | 9662460.0    | 9662460.0    | 9662460.0    
     SAMPLESHEET_CHECK              | memUsage   | 0.15         | 0.15         | 0.15         | 0.15         | 0.15         | 0.15         
     
```

To:

```
     Memory Metrics
    ----------------

     Legend: physical RAM / virtual RAM+swap / %RAM allocated 

     process           | min                | q1                 | median             | q3                 | max
    -------------------+--------------------+--------------------+--------------------+--------------------+--------------------                 
     SAMPLESHEET_CHECK | 9.66M 32.84M 0.15% | 9.66M 32.84M 0.15% | 9.66M 32.84M 0.15% | 9.66M 32.84M 0.15% | 9.66M 32.84M 0.15%    
```

grananda:
> Memory section

Not sure if mixing all metric types in the same cell will help readability and usability. Also, just for memory?

jordeu:
> Not sure if mixing all metric types in the same cell will help readability and usability. Also, just for memory?
It helps in usability because in general you need to check all of then together to decide what parameters to change. Not only for memory. I'll post more examples for the other sections asap.

grananda:
> I'll post more examples for the other sections asap.

Ok, good to me but...this may delay the milestone deadline, can we deal with this in a new issue? I want to ensure I can complete the 2 issues I committed to. 

In addition, we can have a `visualization` option to display the tables in different modes (`extended` and `condensed` views)

jordeu:
> Ok, good to me but...this may delay the milestone deadline, can we deal with this in a new issue? I want to ensure I can complete the 2 issues I committed to.

As you wish. Open the issue and I'll add the other examples there. 


> In addition, we can have a `visualization` option to display the tables in different modes (`extended` and `condensed` views)

I think that there is no need to have two visualizations, I'd go only for the proposed condensed view. In a future we will have the option to output this data as a TSV files (`-o tsv`), so this will be more similar to the `extended` view.

grananda:
New issue created at https://github.com/seqeralabs/tower-cli/issues/116 for table display usability

jordeu (approved):

PR #125, opened 2021-11-15
Rename some commands

### LINKED ISSUE #66: Rename health command to info
### Discussed in https://github.com/seqeralabs/tower-cli/discussions/31

<div type='discussions-op-text'>

<sup>Originally posted by **jordeu** October  7, 2021</sup>
Currently we've the command `health` that it looks like this:
Screenshot from 2021-10-07 11-06-50

I propose to add also an info section something like these:

```
    Details
    -----------------------------+------------------------
     Tower API endpoint          | https://api.tower.nf 
     Tower version               | 21.10.3
     Tower API version           | 1.6.0
     CLI version                 | 1.0.0
     CLI minimum API version     | 1.6.0
     Authenticated user          | jordeu
``` 

I'm proposing to have only one subcommand called `info` with both things (instead of `health` and `info`) trying to keep the number of commands as minimum as possible.
</div>

### LINKED ISSUE #101: Rename participants subcommands `add` to `create` and `change` to `update`
## Description
The `participants` subcommand is the only one using `add` instead of `create` and `change` instead of `update`.  I propose to rename them to use the same names as other commands.

### LINKED ISSUE #107: Replace "create" commands with "add"
Looking at this command 

```
tw credentials create aws
```

Made me think the `create` verb is very confusing, *credentials* are not created but added to the tower config. 

I think the same concept applies to other resources. It would makes sense to replace `create` verb into `add`

jordeu:
-  Rename participants change to participants update. Fixes #101 
-  Rename all 'create' commands to 'add' command. Fixes #107 
-  Rename health command to info and add extra fields. Resolves #66


grananda (approved):

swampie (commented):
A small change

PR #133, opened 2021-11-16
Use parametrized test to test all output formats

### LINKED ISSUE #132: Automatically test JSON output
# Description

Use parametrized tests to test always all output formats.

jordeu:
Resolves #132


src/main/java/io/seqera/tower/cli/commands/workspaces/ListCmd.java
```diff
@@ -38,15 +38,15 @@
...
-
-            return new WorkspaceList(userName(), responseOrg);
+        if (response.getOrgsAndWorkspaces() == null) {
+            return new WorkspaceList(userName(), Collections.emptyList());
         }
 
-        return new WorkspaceList(userName(), response.getOrgsAndWorkspaces());
+        List<OrgAndWorkspaceDbDto> responseOrg = response.getOrgsAndWorkspaces()
+                .stream()
+                .filter(v -> v.getWorkspaceId() != null && (organizationName == null || organizationName.equals(v.getOrgName()))
```
grananda, line 47: What is purpose of the  `organizationName == null` condition?

jordeu: When `organizationName` is `null` we don't want to apply this filter `organizationName.equals(v.getOrgName())` but otherwise we want to apply it (this is why I'm using the OR). The new thing here is to filter out the ones that have a `null` `v.getWorkspaceId()`.

grananda (approved):
I dont see anything out of place. Test run great. Just that comment I made before.

src/main/java/io/seqera/tower/cli/commands/workspaces/ListCmd.java
grananda: I just can't imagine a scenario with a nullable organization name. Was wondering whether if it was redundant.

jordeu: Si que puede ser `null`. Cuando haces un `tw workspaces list` sin el `-o` que es opcional, te devuelve una lista con todos los workspaces en todas las orgs.

swampie (approved):

PR #153, opened 2021-11-22
Task 63 / Allow workspace reference

grananda:
Allow the user of organization/workspace name or workspace id to identify the command workspace


src/main/java/io/seqera/tower/cli/commands/global/WorkspaceOptionalOptions.java
```diff
@@ -18,5 +18,5 @@
     public static final String DEFAULT_VALUE = "${TOWER_WORKSPACE_ID}";
 
     @CommandLine.Option(names = {"-w", "--workspace"}, description = DESCRIPTION, defaultValue = DEFAULT_VALUE)
```
jordeu, line 20: Update the description to reflect this change

jordeu (changes requested):

jordeu (approved):

PR #157, opened 2021-11-29
Task 35 / Datasets

grananda:


grananda:
@jordeu and @swampie this is mostly finished. Still, the upload endpoint has to be fixed. There seems to be a mismatch between the expected and the sent mime type.

````
nov. 29, 2021 11:52:46 A. M. org.glassfish.jersey.logging.LoggingInterceptor log
INFO: 2 * Sending client request on thread main
2 > POST http://localhost:8080/workspaces/249664655368293/datasets/1KIYOm3Bbdxf9Vczf0Vdaz/upload?header=false
2 > Accept: application/json
2 > Authorization: Bearer eyJ0aWQiOiAyfS5iODcxMGUzYzFhYmUwNjUzODlmOGJlNjE4MzAzNGQyZGU3MTE3ZDU5
2 > Content-Type: multipart/form-data
2 > User-Agent: OpenAPI-Generator/1.2.0-SNAPSHOT/java
--Boundary_1_794563378_1638183166414
Content-Type: application/octet-stream
Content-Disposition: form-data; filename="samplesheet_test.csv"; size=1121; name="file"

sample,fastq_1,fastq_2,strandedness
WT_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357070_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357070_2.fastq.gz,reverse
WT_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357071_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357071_2.fastq.gz,reverse
WT_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357072_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357072_2.fastq.gz,reverse
RAP1_UNINDUCED_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357073_1.fastq.gz,,reverse
RAP1_UNINDUCED_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357074_1.fastq.gz,,reverse
RAP1_UNINDUCED_REP2,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357075_1.fastq.gz,,reverse
RAP1_IAA_30M_REP1,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357076_1.fastq.gz,s3://nf-core-awsmegatests/rnaseq/input_data/minimal/GSE110004/SRR6357076_2.fastq.gz,reverse

--Boundary_1_794563378_1638183166414--


nov. 29, 2021 11:52:46 A. M. org.glassfish.jersey.logging.LoggingInterceptor log
INFO: 2 * Client response received on thread main
2 < 400
2 < connection: keep-alive
2 < content-length: 149
2 < Content-Type: application/json
2 < date: Mon, 29 Nov 2021 10:52:46 GMT
{"message":"Given file is not a dataset file. Detected media type: 'application/octet-stream'. Allowed types: 'text/csv, text/tab-separated-values'"}
```

There are also some config reflection issues to be resolved as well.

jordeu:
It seems that the OpenAPI code generation has some limitations and it is not possible to set the content-type of the file that you upload. I've opened an issue [here](https://github.com/seqeralabs/tower-sdk-gencode/issues/1) to investigate further when we've some time.

Meanwhile, I've solved this at CLI side with this workaround that overrides an API client method:
https://github.com/seqeralabs/tower-cli/blob/d0f68a0264a8f649a85d7c99a3f611b9a34bd4f4/src/main/java/io/seqera/tower/cli/commands/AbstractApiCmd.java#L108-L122

build.gradle
```diff
@@ -8,15 +8,18 @@ plugins {
 
 repositories {
     mavenCentral()
-    maven { url = "https://s3-eu-west-1.amazonaws.com/maven.seqera.io/releases" }
+//    maven { url = "https://s3-eu-west-1.amazonaws.com/maven.seqera.io/releases" }
+    maven { url = "https://s3-eu-west-1.amazonaws.com/maven.seqera.io/snapshots" }
```
jordeu, line 12: I've released SDK version 1.2.0. Rollback to use releases URLs.

build.gradle
```diff
@@ -8,15 +8,18 @@ plugins {
...
 }
 
 dependencies {
     implementation 'javax.activation:activation:1.1.1'
     implementation 'org.slf4j:slf4j-api:1.7.32'
     implementation 'ch.qos.logback:logback-core:1.2.6'
     implementation 'ch.qos.logback:logback-classic:1.2.6'
-    implementation 'io.seqera:tower-java-sdk:1.1.0'
+//    implementation 'io.seqera:tower-java-sdk:1.1.0'
+    implementation 'io.seqera:tower-java-sdk:1.2.0-SNAPSHOT'
```
jordeu, line 22: Use new 1.2.0 release

samplesheet_test.csv
```diff
@@ -0,0 +1,8 @@
+sample,fastq_1,fastq_2,strandedness
```
jordeu, line 1: Move this file into a test folder (or remove it).

src/main/java/io/seqera/tower/cli/commands/datasets/CreateCmd.java
```diff
@@ -0,0 +1,59 @@
...
+    public String description;
+
+    @CommandLine.Option(names = {"--header"}, description = "Set first row as a header.")
+    public boolean header = false;
+
+    @CommandLine.Parameters(index = "0", paramLabel = "FILENAME", description = "Data file to upload.", arity = "1")
+    Path fileName = null;
+
+    @CommandLine.Mixin
+    public WorkspaceOptionalOptions workspace;
```
jordeu, line 44: Workspace must be required because datasets it is not available at personal workspace.

src/main/java/io/seqera/tower/cli/utils/FilesHelper.java
```diff
@@ -39,4 +40,19 @@ public static void saveString(String fileName, String text) {
             e.printStackTrace();
         }
     }
+
+    public static String readFile(File file) {
```
jordeu, line 44: This method reads a file and deletes it always. Rename the method name to make it clear. Maybe better `readFileAndDelete` ?

jordeu (changes requested):
Some binary tests are failing

src/main/java/io/seqera/tower/cli/commands/datasets/CreateCmd.java
grananda: I am assuming here that the user will pull from the `env` variable for workspace settings. If I force the workspace option on dataset commands, the `env` will be ignored. On the other hand, not having such a variable will lead to an error. It could be confusing. A mandatory default profile system is becoming more necessary as we evolve.

jordeu: If the variable is available then the picocli `required` flag it is not going to give error and use it. And if the user do not give a `-w` or sets the `env` variable then the command has to fail with a required `workspace` error.

src/main/java/io/seqera/tower/cli/commands/datasets/CreateCmd.java
```diff
@@ -0,0 +1,59 @@
...
+import io.seqera.tower.cli.responses.datasets.DatasetCreate;
+import io.seqera.tower.model.CreateDatasetRequest;
+import io.seqera.tower.model.CreateDatasetResponse;
+import picocli.CommandLine;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+@CommandLine.Command(
+        name = "create",
```
jordeu, line 26: Rename the command and the class to `add` as we did with all the other `create` subcommands that we had.

grananda:
> Meanwhile, I've solved this at CLI side with this workaround that overrides an API client method:

Not sure this solves the issue. Have you tried it?

jordeu:
> > Meanwhile, I've solved this at CLI side with this workaround that overrides an API client method:
> 
> Not sure this solves the issue. Have you tried it?

Yes, I was able to create a dataset and upload a CSV file.

jordeu:
Currently there is only one binary test failing on Linux. It is a problem of the static linked image generation, locales are missing at MUSL when doing the image compilation. Adding the debian package locales does not fix the problem.

src/main/java/io/seqera/tower/cli/commands/datasets/UpdateCmd.java
```diff
@@ -0,0 +1,65 @@
...
+    @CommandLine.Mixin
+    public WorkspaceRequiredOptions workspace;
+
+    @Override
+    protected Response exec() throws ApiException, IOException {
+        Long wspId = workspaceId(workspace.workspace);
+        Dataset dataset = fetchDescribeDatasetResponse(datasetRefOptions, wspId);
+        UpdateDatasetRequest request = new UpdateDatasetRequest();
+        request.setName(newName != null ? newName : dataset.getName());
+        request.setDescription(description);
```
jordeu, line 55: If `description` is `null` previous description is overwritten

src/main/java/io/seqera/tower/cli/commands/datasets/UpdateCmd.java
```diff
@@ -0,0 +1,65 @@
...
+
+        api().updateDataset(wspId, dataset.getId(), request);
+
+        if (fileName != null) {
```
jordeu, line 59: We need to give an error somewhere if the user adds the option `--header` but not the `--file`.

jordeu (changes requested):
I've removed non-ASCII characters from the tests to avoid binary tests to fail. I'll open an issue to track that problem.

grananda:
> I've removed non-ASCII characters from the tests to avoid binary tests to fail. I'll open an issue to track that problem.

Really?! I looked like crazy for weird stuff in the response classes yesterday, would never imagine the error came from there. Thanks!

jordeu:
Related issue https://github.com/seqeralabs/tower-cli/issues/161

jordeu (approved):

PR #167, opened 2021-12-02
Task 116 / feature: Reformat metric tables

grananda:


grananda:
@jordeu and @swampie this is currently working but the test is failing because it is parsing the JSON as `int` and `double` rather than the float we need`float`.

grananda:
In addition to the comment above, I still need to make some adjustments to the table to make a better fit for missing content

image

src/main/java/io/seqera/tower/cli/responses/runs/RunViewMetrics.java
```diff
@@ -92,35 +187,79 @@ private void processDataTable(List<Map<String, Object>> metricData, PrintWriter
...
+     * @return
+     */
+    private Function<Float, Object> getBlockPrettyTransformation(String blockKey, Map<String, Function<Float, Object>> formatter) {
+        return formatter
+                .entrySet()
+                .stream()
+                .filter(fmt -> Objects.equals(fmt.getKey(), blockKey))
+                .findFirst()
+                .orElse(null)
+                .getValue();
```
jordeu, line 263: Possible null pointer exception here.

src/main/java/io/seqera/tower/cli/responses/runs/RunViewMetrics.java
```diff
@@ -92,35 +187,79 @@ private void processDataTable(List<Map<String, Object>> metricData, PrintWriter
...
+
+    /**
+     * Find the right pretty transformation for the given data block and a transformation map.
+     *
+     * @param blockKey
+     * @param formatter
+     * @return
+     */
+    private Function<Float, Object> getBlockPrettyTransformation(String blockKey, Map<String, Function<Float, Object>> formatter) {
+        return formatter
```
jordeu, line 257: Is this equivalent to a simply map get?
 
```
return formatter.getOrDefault(blockKey, null);
```

src/main/java/io/seqera/tower/cli/responses/runs/RunViewMetrics.java
```diff
@@ -67,7 +75,90 @@ public Object getJSON() {
...
+        cols.add("process");
+        if (groupType == MetricPreviewFormat.expanded) {
+            cols.add("metric");
+        }
+        cols.addAll(fields);
+
+        if (!metricsMem.isEmpty()) {
+            out.println(ansi(String.format("%n%n    @|bold  Memory Metrics|@%n    ----------------%n")));
+
+            Map<String, Function<Float, Object>> formatter = new HashMap<>();
```
jordeu, line 91: Maybe you can use `Number` instead of `Float` to avoid the test failures.

src/main/java/io/seqera/tower/cli/responses/runs/RunViewMetrics.java
```diff
@@ -67,7 +75,90 @@ public Object getJSON() {
...
+            } else {
+                processExpandedDataTable(metricsMem, out, cols, formatter);
+            }
+        }
+
+
+        if (!metricsCpu.isEmpty()) {
+            out.println(ansi(String.format("%n%n    @|bold  CPU Metrics|@%n    ----------------%n")));
+
+            Map<String, Function<Float, Object>> formatter = new HashMap<>();
```
jordeu, line 108: Why not create only one static formatter with all the formatters?  You can create a static map or a helper class `formatters.get(String sectionName) -> Function<Number, Object>`

src/main/java/io/seqera/tower/cli/responses/runs/RunViewMetrics.java
```diff
@@ -92,35 +187,79 @@ private void processDataTable(List<Map<String, Object>> metricData, PrintWriter
...
+        metricData.forEach(processDataBlock -> {
+            processDataBlock.forEach((process, sectionDataBlock) -> {
+                List<String> cells = new ArrayList<>();
+                cells.add(process);
+                Map<String, List<String>> data = summarizeDataBlocks((Map<String, Map<String, Float>>) sectionDataBlock, formatter);
+                if (data.size() > 0) {
 
+                    // This where summarized data cells are created into a concatenated string.
+                    data.values().stream().forEach(it -> {
+                        cells.add(String.join(" / ", it));
```
jordeu, line 211: Given that we want a condensed view, I think that is better to use just a `" "` blank space to separate the different values.

jordeu (changes requested):

src/main/java/io/seqera/tower/cli/responses/runs/utils/MetricFormatMapper.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+import io.seqera.tower.cli.utils.FormatHelper;
+
+public class MetricFormatMapper {
+
+    public static Map<String, Function<Number, Object>> getMap() {
```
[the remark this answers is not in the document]
jordeu: Also add a method to avoid always using it like `.getMap().getOrDefault(dataBlockDef, null)`.

src/main/java/io/seqera/tower/cli/responses/runs/RunViewMetrics.java
```diff
@@ -92,35 +170,60 @@ private void processDataTable(List<Map<String, Object>> metricData, PrintWriter
...
+        data.entrySet().stream().forEach(it -> {
+            if (it.getValue() != null) {
+                Function<Number, Object> fnc = MetricFormatMapper.getMap().getOrDefault(it.getKey(), null);
+
+                for (Map.Entry<String, Number> entry : it.getValue().entrySet()) {
+                    if (!result.containsKey(entry.getKey())) {
+                        result.put(entry.getKey(), new ArrayList<>());
+                    }
 
+                    result.get(entry.getKey()).add(fnc.apply(entry.getValue()).toString());
```
jordeu, line 222: Add some padding here to align the table results:
```suggestion
                    result.get(entry.getKey()).add(String.format("%1$9s", fnc.apply(entry.getValue())));
```
Or even better, inside each formatter because then you have a better idea of the padding length.

jordeu (changes requested):
Some formatting changes to make it more readable the condensed view:

- Instead of `31.46 MB 5.05 MB 0.01%` better `31MB 5MB 0%` without decimals.
- Instead of `44s 3m  0.39%` better `3m44s  0%` notice that minutes go before seconds.

grananda:
@jordeu last comment was done

src/main/java/io/seqera/tower/cli/responses/runs/utils/MetricFormatMapper.java
```diff
@@ -0,0 +1,69 @@
...
+
+        return instance;
+    }
+
+    public Function<Number, Object> getFormatTransformer(String key) {
+        return formatter().getOrDefault(key, null);
+    }
+
+    public Number getPadding(String key) {
+        return padding().getOrDefault(key, null);
```
jordeu, line 40: The same here, you're always creating a new map instance.

jordeu (changes requested):

src/main/java/io/seqera/tower/cli/responses/runs/RunViewMetrics.java
```diff
@@ -67,7 +76,72 @@ public Object getJSON() {
...
+                processExpandedDataTable(metricsCpu, out, cols);
+            }
+        }
+
+
+        if (!metricsTime.isEmpty()) {
+            out.println(ansi(String.format("%n%n    @|bold  Time Metrics|@%n    ----------------%n")));
+
+            if (groupType == MetricPreviewFormat.condensed) {
+                out.println(ansi(String.format("   @|italic   Legend: reads / writes|@%n")));
```
jordeu, line 117: This legend it is not correct.

src/main/java/io/seqera/tower/cli/responses/runs/utils/MetricFormatMapper.java
```diff
@@ -0,0 +1,69 @@
...
+    public static MetricFormatMapper getInstance() {
+        if (instance == null) {
+            instance = new MetricFormatMapper();
+        }
+
+        return instance;
+    }
+
+    public Function<Number, Object> getFormatTransformer(String key) {
+        return formatter().getOrDefault(key, null);
```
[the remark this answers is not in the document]
grananda: Not sure is the latest solution is the singleton you expect. Please check.

jordeu:
Nice result!

Screenshot from 2021-12-03 17-35-32

PR #178, opened 2021-12-03
Task 105 / feature: set visibility when creating a workspace

grananda:


jordeu:
This seems correct to me. But we will wait until shared workspaces are available at `tower.nf`.

Issue #177, opened 2021-12-03
Command to watch / tail logs from a run


ewels:
It would be super helpful if the `tw` cli client could have a command / flag to watch the execution of a run and exit only when the run completes.

### Suggested implementation

* `tw runs watch --id xxx`
* `tw launch --watch`

This would then initially show the changing status of the run (initialising / running etc) and then tail the main Nextflow log as it is produced. When the run completes the cli should exit, using exit code `0` if the run was successful and non-zero if there was an error.

### Use case

The use case I have in mind is for GitHub actions CI. We [now use](https://github.com/nf-core/tower-action) `tw` to launch test workflows on AWS but it is super difficult to feed back the results to the CI. I'm also thinking about [more complicated methods](https://github.com/seqeralabs/nf-tower-cloud/issues/2358) to achieve a similar end result (which could still be good for longer-running workflows?) but a far simpler approach would be to keep the CI job that launched the workflow running until it is complete.

Getting the Nextflow log output would be a bonus, as we would be able to see any errors directly in the GitHub Actions run without having to go into Tower (especially good for @nf-core as most people running workflows won't have an account in Tower).

The only other way I can think of is to collect the run ID after launch and then have a bash loop running `tw runs list` that checks the status of the run until it completes. But this is not nearly as elegant.

BrunoGrandePhD:
I think this feature would unlock some cool automation. For example, at Sage, we currently need to run a [workflow](https://github.com/Sage-Bionetworks-Workflows/nf-synstage) beforehand to stage the data from Synapse (a controlled-access data repository) and another [workflow](https://github.com/Sage-Bionetworks-Workflows/nf-synindex) afterwards to index the output files back into Synapse. 

We would like to automate this process using the new Tower CLI, but we need the ability to know when a workflow is done running. Phil's idea of a bash loop also came to mind, but like him, I think it's a hacky solution.

pditommaso:
I agree it'd make sense to implement this asap

PR #216, opened 2022-04-11
Allow to wait for a given status after creating a compute environment and launching a workflow.

jordeu:
# Description

Added new `--wait` flag that waits until a given status before exit. 

Added to commands:
- `tw compute-envs add ...`
- `tw compute-envs import ...`
- `tw launch ...`

Exit code:
- It return with a 0 exit code if the entity reaches the desired status. 
- It returns an error 1 exit code if the entity reaches a status that is incompatible with the desired status. (for example if a workflow desired status is 'RUNNING' and the job gets status 'FAILED' or 'CANCELLED' before).

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/AbstractApiCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/AbstractApiCmd.java
@@ -15,6 +15,7 @@ import io.seqera.tower.ApiClient;
 import io.seqera.tower.ApiException;
 import io.seqera.tower.api.DefaultApi;
 import io.seqera.tower.cli.Tower;
+import io.seqera.tower.cli.commands.enums.OutputType;
 import io.seqera.tower.cli.exceptions.ComputeEnvNotFoundException;
 import io.seqera.tower.cli.exceptions.NoComputeEnvironmentException;
 import io.seqera.tower.cli.exceptions.OrganizationNotFoundException;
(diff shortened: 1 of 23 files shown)


pditommaso:
How the desired status is specified?

ewels:
Looks like this should also close https://github.com/seqeralabs/tower-cli/issues/177 👏🏻 🚀

jordeu:
> How the desired status is specified?

Basically `--wait=<STATUS>`. I've updated the description with an usage example.

pditommaso (approved):

README.md
```diff
@@ -144,6 +144,8 @@ Pipelines consist of a pipeline repository, launch parameters, and a Compute Env
 tw launch my_sleepy_pipeline
 ```
 
+> Add a `--wait=SUCCEEDED` if you want the command to wait until the pipeline execution is complete.
```
ewels, line 147: Might be nice to have some mention of the different exit codes as well (and under which conditions they occur) 👍🏻

pditommaso: I think the error code are only 0 (ok) and 1 (fail)

ewels: Yup, but that's useful to know! 😉 

I was also slightly unclear on the nitty gritty. For example, if I do `--wait=FAILED` and the run finishes with `SUCCEEDED` will `tw` effectively hang, waiting forever? Or exit with `1`? Or exit with `0` because it was a successful run?

ewels:
```suggestion
Add a `--wait=SUCCEEDED` if you want the command to wait until the pipeline execution is complete.

When using `--wait`, `tw` can exit with one of two exit codes:

- `0`: When the run/compute-env reaches the desired state.
- `1`:  When the run/compute-env reaches a state that makes it impossible to reach the desired state.
```

jordeu: hang waiting forever (until you press Ctrl+C) :laughing:

jordeu: Ok, let me fix these corner cases. I'm going to return a 1 exit code, because I imagine that someone does a `--wait=FAILED` because he wants to test that the run fails (like a unit testing of a pipeline).

Issue #220, opened 2022-04-12
Flag `--no-max` doesn't work as expected


ewels:
When listing members of an organization, the flag `--no-max` has the opposite effect of what was expected: it shows _fewer_ records instead of showing all of them. Just 10 records, specifically. The organization currently has 24 members (which is what is shown without the `--no-max` flag).

Example below (details redacted):

```console
 ❯ tw members list -o=NGI

  Members for NGI organization:

     ID              | Username          | Email                           | Role
    -----------------+-------------------+---------------------------------+--------
     xxxxxxxxxxxxxxx | xxxxx             | xxxxx@xxxxxxxxxx.xx             | OWNER
     xxxxxxxxxxxxxxx | xxxxxx-xxxxxxxxx  | xxxxxx.xxxxxxxxx@xxxxxxxxxx.xx  | MEMBER
     xxxxxxxxxxxxxxx | xxxxxxxxxxx       | xxxxxxxxxxx@xxxxx.xxx           | MEMBER
     xxxxxxxxxxxxxxx | xxxxx-xxxx        | xxxxx.xxxx@xxxxxxxxxx.xx        | MEMBER
     xxxxxxxxxxxxxxx | xxxxx-xxxxxxxx    | xxxxx.xxxxxxxx@xxxxxxxxxx.xx    | MEMBER
     xxxxxxxxxxxxxxx | xxxx-xxxxxxxxx    | xxxx.xxxxxxxxx@xxxx.xx          | MEMBER
     xxxxxxxxxxxxxx  | xxxxyxxxx-xxxx    | xxxxyxxxx.xxxx@xxxxxxxxxx.xx    | MEMBER
     xxxxxxxxxxxxxxx | xxxxx             | xxxx.xxxxx@xxxxxxxxxx.xx        | OWNER
     xxxxxxxxxxxxxxx | xxxxy-xxxxxxxx    | xxxxy.xxxxxxxx@xxxxxxxxxx.xx    | MEMBER
     xxxxxxxxxxxxxxx | xxxxzxxxx-xxxxxx  | xxxxzxxxx.xxxxxx@xxxxxxxxxx.xx  | MEMBER
     xxxxxxxxxxxxxxx | xxxxxx            | xxxxxx@xxxx.xxxxxxxx-zxxxxx.xx  | MEMBER
     xxxxxxxxxxxxxxx | xxxxx-xxxxxxx     | xxxxx.xxxxxxx@xxx.xx.xx         | OWNER
[...]

PR #221, opened 2022-04-12
Remove --no-max pagination option

jordeu:
## Description

Removing the option `--no-max` because it is not working as expected. The API cannot return all the elements of a list, so we need to implement pagination multi requests at CLI side. Related issue #220

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/global/PaginationOptions.java
+++ b/src/main/java/io/seqera/tower/cli/commands/global/PaginationOptions.java
@@ -37,20 +37,13 @@ public class PaginationOptions {
     public static class Sizeable {
         @CommandLine.Option(names = {"--max"}, description = "Maximum number of records to display [default: " + MAX + "].")
         public Integer max;
-
-        @CommandLine.Option(names = {"--no-max"}, description = "Show all records.")
-        public Boolean noMax;
     }
 
     public static Integer getMax(PaginationOptions paginationOptions) throws TowerException {
         Integer max = PaginationOptions.MAX;
 
         if (paginationOptions.sizeable != null) {
-            if (paginationOptions.sizeable.noMax != null && paginationOptions.sizeable.max != null) {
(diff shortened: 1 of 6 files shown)


swampie (approved):
Is it something we have to fix on the backend? @jordeu

jordeu:
> Is it something we have to fix on the backend? @jordeu

No

src/main/java/io/seqera/tower/cli/commands/global/PaginationOptions.java
```diff
@@ -37,20 +37,13 @@ public static class Pageable {
...
     public static Integer getMax(PaginationOptions paginationOptions) throws TowerException {
         Integer max = PaginationOptions.MAX;
 
         if (paginationOptions.sizeable != null) {
-            if (paginationOptions.sizeable.noMax != null && paginationOptions.sizeable.max != null) {
-                throw new TowerException("Please use either --no-max or --max as pagination size parameter");
-            }
-
-            max = paginationOptions.sizeable.noMax != null ? null : paginationOptions.sizeable.max != null ? paginationOptions.sizeable.max : max;
+            max = paginationOptions.sizeable.max != null ? paginationOptions.sizeable.max : max;
```
pditommaso, line 46: What's preventing the user to provide a stupidly large number for max?

jordeu: Nothing at the CLI. The backend has a maximum maximum, so even if they request for a big number they will get only the backend maximum. This backend maximum is configurable at backend side so we do not known it a priori at CLI side and cannot do a warning or error

pditommaso: ok

jordeu: Somehow related to this I've opened this issue: https://github.com/seqeralabs/tower-cli/issues/223

PR #241, opened 2022-08-01
Never set the sourceWorkspaceId on a relaunch

jordeu:
## Description
- Never set the sourceWorkspaceId on a relaunch
- Fix invalid error message at info subcommand


VERSION
```diff
@@ -1 +1 @@
-0.6.2
+0.6.3
```
pditommaso, line 1: I'd better to not include the version in the PR

jordeu: Normally I create a PR only to change the version and release. This is a quick bug fix release and I was a bit lazy to create another PR. But ok I'll do PRs even for small bug fix releases.

pditommaso: no pb, I can approve this. but IMO version should not go into the PR (imagine to cherry picking into another branch). The version can be pushed directly to build the final release

pditommaso (approved):

Issue #230, opened 2022-05-26
Add labels support


jordeu:
## Description

Add support to manage and add labels to `actions`, `pipelines`, `runs` and `compute-envs`.

### Label management

**Create**
```
tw labels add [OPTIONS]
   * -n, --name=<name>              Label name
     -w, --workspace=<workspace>    Workspace numeric identifier (TOWER_WORKSPACE_ID as default)
   -v, --value=<label value>
``` 

**Update**
``` 
tw labels update [OPTIONS]
    * -i, --id=<labelId>             Label identifier
      -w, --workspace=<workspace>    Workspace numeric identifier (TOWER_WORKSPACE_ID as default)
     -v, --value=<label value>
``` 

**Delete**
```
tw labels delete [OPTIONS]
   * -i, --id=<labelId>             Label identifier
     -w, --workspace=<workspace>    Workspace numeric identifier (TOWER_WORKSPACE_ID as default)
```

**List**
```
tw labels list [OPTIONS]
     * -w, --workspace=<workspace>   Workspace numeric identifier (TOWER_WORKSPACE_ID as default)
     -t, --type=<type>         Type of Label
     -f, --filter=<filter>         Show only labels that contain the given word.
         --page=<page>             Pages to display [default: 1].
         --offset=<offset>         Rows record offset [default: 0].
         --max=<max>               Maximum number of records to display [default: 100].
```

### Show labels

**Actions**

Add a labels section
```
tw actions view -i <action_id> --show-labels
```

Add a labels column
```
tw actions list --show-labels
```

**Pipelines**
``` 
tw pipelines view -i <pipeline_id> --show-labels
tw pipelines list --show-labels
```

**Runs**
```
tw runs view -i <id> --show-labels
tw runs list --show-labels
```

**Compute environments**
``` 
tw compute-envs view -i <id> --show-labels
tw compute-envs list --show-labels
```

### Manage labels on entities

```
tw [actions|pipelines|runs] labels -w () -i () -n () key1=value1,key2=value2 --operation=[set,append,delete] --no-create
```

### Set labels on entities add subcommand
```
tw [compute-envs|actions|pipelines] add ... --labels key1=valu1,...
```

## Doubts / Missing things

- Allow to filter lists by label. I propose to wait that it is possible to filter using the "search" syntax.
- Do we allow to import/export labels when using the import/export commands? 
- Allow to add labels when creating/updating resource? Or force to use always the attach/detach commands?

## Tasks

[...]

PR #297, opened 2023-03-16
Label management commands

JaimeSeqLabs:
See #230

Add CRUD commands for labels.

Work in progress

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/labels/AddLabelsCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/labels/AddLabelsCmd.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2021, Seqera Labs.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ * This Source Code Form is "Incompatible With Secondary Licenses", as
+ * defined by the Mozilla Public License, v. 2.0.
+ */
+
+package io.seqera.tower.cli.commands.labels;
+
+import io.seqera.tower.ApiException;
+import io.seqera.tower.cli.commands.computeenvs.AbstractComputeEnvCmd;
+import io.seqera.tower.cli.commands.global.WorkspaceOptionalOptions;
+import io.seqera.tower.cli.exceptions.ShowUsageException;
+import io.seqera.tower.cli.exceptions.TowerException;
+import io.seqera.tower.cli.responses.Response;
+import io.seqera.tower.model.CreateLabelRequest;
+import io.seqera.tower.model.CreateLabelResponse;
+import picocli.CommandLine;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+@CommandLine.Command(
+        name = "add",
+        description = "Add new label"
+)
+public class AddLabelsCmd extends AbstractLabelsCmd {
+
+    @CommandLine.Option(names = {"-n", "--name"}, description = "Label name.", required = true)
+    public String labelName;
+
+    @CommandLine.Option(names = {"-w", "--workspace"}, description = "Workspace numeric identifier", required = false)
+    @Nullable
+    public Long workspaceId;
+
+    @CommandLine.Option(names = {"-v", "--value"}, description = "label value", required = false)
+    @Nullable
+    public String labelValue;
+
+    @Override
+    protected Response exec() throws ApiException, IOException, IllegalArgumentException, TowerException {
+
+        if (labelName == null || labelName.isEmpty()) {
+            throw new IllegalArgumentException("Missing label name");
+        }
+
+        /*
+        System.out.println("############################");
+
+        System.out.println(labelName);
+        System.out.println(labelValue);
+        System.out.println(workspace.workspace);
+        */
+
+        CreateLabelRequest req = new CreateLabelRequest();
+
+        req.setName(labelName);
+
+        if (labelValue != null) {
+            req.setValue(labelValue);
+            req.setResource(true);
+        } else {
+            req.setResource(false);
+        }
+
+
+        CreateLabelResponse res = null;
+        try {
+            res = api().createLabel(req, workspaceId);
+        } catch (Exception e) {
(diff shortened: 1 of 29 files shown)


src/main/java/io/seqera/tower/cli/commands/labels/AddLabelsCmd.java
```diff
@@ -0,0 +1,73 @@
...
+@CommandLine.Command(
+        name = "add",
+        description = "Add new label"
+)
+public class AddLabelsCmd extends AbstractLabelsCmd {
+
+    @CommandLine.Option(names = {"-n", "--name"}, description = "Label name.", required = true)
+    public String labelName;
+
+    @CommandLine.Option(names = {"-w", "--workspace"}, description = "Workspace numeric identifier", required = false)
```
jordeu, line 37: All the descriptions finish with a dot.

src/main/java/io/seqera/tower/cli/commands/labels/AddLabelsCmd.java
```diff
@@ -0,0 +1,73 @@
...
+    @Nullable
+    public Long workspaceId;
+
+    @CommandLine.Option(names = {"-v", "--value"}, description = "label value", required = false)
```
jordeu, line 41: And all the descriptions start with uppercase

src/main/java/io/seqera/tower/cli/commands/labels/AddLabelsCmd.java
```diff
@@ -0,0 +1,73 @@
...
+        description = "Add new label"
+)
+public class AddLabelsCmd extends AbstractLabelsCmd {
+
+    @CommandLine.Option(names = {"-n", "--name"}, description = "Label name.", required = true)
+    public String labelName;
+
+    @CommandLine.Option(names = {"-w", "--workspace"}, description = "Workspace numeric identifier", required = false)
+    @Nullable
+    public Long workspaceId;
```
jordeu, line 39: Use `WorkspaceOptionalOptions` mixin and then the method `workspaceId` to resolve the workspace, this allows to use same parameter to pass the workspace using the name or the id.

See a usage example here:
http://github.com/seqeralabs/tower-cli/blob/70c30c586a05cfc44819aa0cfd2204291f95e7b4/src/main/java/io/seqera/tower/cli/commands/secrets/AddCmd.java#L43-L43

src/main/java/io/seqera/tower/cli/commands/labels/DeleteLabelsCmd.java
```diff
@@ -0,0 +1,49 @@
...
+    protected Response exec() throws ApiException, IOException {
+        try {
+
+            api().deleteLabel(labelId, workspaceId);
+
+        } catch (Exception e) {
+            throw new TowerException(String.format("Unable to delete label '%d', reason: %s", labelId, e.toString()));
+        }
+
+        return new GenericStrResponse(String.format("Label '%d' deleted successfully\n", labelId));
```
jordeu, line 47: Remember that with the option `--output json` the response will be serialized to JSON and show as result. This is why a generic response is not a good idea

See a similar example here http://github.com/seqeralabs/tower-cli/blob/70c30c586a05cfc44819aa0cfd2204291f95e7b4/src/main/java/io/seqera/tower/cli/responses/secrets/SecretDeleted.java#L17-L17

src/main/java/io/seqera/tower/cli/commands/labels/ListLabelsCmd.java
```diff
@@ -0,0 +1,34 @@
...
+import io.seqera.tower.cli.responses.Response;
+import picocli.CommandLine;
+
+import java.io.IOException;
+
+@CommandLine.Command(
+        name = "list",
+        description = "List labels"
+)
+public class ListLabelsCmd extends AbstractComputeEnvCmd {
```
jordeu, line 28: All other subcommands do not have the parent command nomenclature at the name. So here would be `ListCmd` instead of `ListLabelsCmd`. I do not have a strong opinion on what is better, just right now all the others are like that.

src/main/java/io/seqera/tower/cli/commands/labels/DeleteLabelsCmd.java
```diff
@@ -0,0 +1,49 @@
...
+    public Long workspaceId;
+
+    @Override
+    protected Response exec() throws ApiException, IOException {
+        try {
+
+            api().deleteLabel(labelId, workspaceId);
+
+        } catch (Exception e) {
+            throw new TowerException(String.format("Unable to delete label '%d', reason: %s", labelId, e.toString()));
```
jordeu, line 44: If there is no good reason and the error message that comes from the API is good enough we tend to manage all the exceptions in an unify way here:
http://github.com/seqeralabs/tower-cli/blob/70c30c586a05cfc44819aa0cfd2204291f95e7b4/src/main/java/io/seqera/tower/cli/utils/ResponseHelper.java#L79-L94

So there is no need to try-catch in most of the commands.

src/main/java/io/seqera/tower/cli/commands/labels/AddLabelsCmd.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @CommandLine.Option(names = {"-v", "--value"}, description = "label value", required = false)
+    @Nullable
+    public String labelValue;
+
+    @Override
+    protected Response exec() throws ApiException, IOException, IllegalArgumentException, TowerException {
+
+        if (labelName == null || labelName.isEmpty()) {
```
jordeu, line 48: no need to double check this, if the parameter is required picocli already checks this

src/main/java/io/seqera/tower/cli/commands/labels/AddLabelsCmd.java
```diff
@@ -0,0 +1,73 @@
...
+            throw new IllegalArgumentException("Missing label name");
+        }
+
+        CreateLabelRequest req = new CreateLabelRequest();
```
jordeu, line 52: when possible is nicer to use the builder pattern `new CreateLabelRequest().name(labelName).value(labelValue).resource(labelValue != null)`

jordeu:
Thanks @JaimeSeqLabs! let's fix these small comments and then merge it and I'll take care of do the testing in the other branch altogether with other label features.

PR #300, opened 2023-03-17
Show labels pull request

pabloaledo:
See #230 .

Add --show-labels to actions, pipelines, runs, and compute environments.

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/runs/ViewCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/runs/ViewCmd.java
@@ -37,6 +37,8 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
+import static io.seqera.tower.cli.utils.FormatHelper.formatLabels;
+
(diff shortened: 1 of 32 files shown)


src/main/java/io/seqera/tower/cli/commands/runs/ViewCmd.java
```diff
@@ -149,6 +161,21 @@ protected Response exec() throws ApiException {
                 utilization.put("cpuEfficiency", progress.getWorkflowProgress().getCpuEfficiency());
             }
 
+            if (labels != null && !labels.getLabels().isEmpty()) {
+                general.put(
+                    "labels",
+                    labels.getLabels().stream()
```
jordeu, line 167: This logic to convert a list of labels into a string is duplicated in several places. Maybe is better to move it into a shared method at `FormatHelper` class.

jordeu:
Thanks @pabloaledo! let's fix these small comments and the failing tests and then merge it. I'll take care of do the testing and missing things in the other branch altogether with other label features.

jordeu (approved):

PR #333, opened 2023-09-14
Add overwrite option when importing pipelines and ces

### LINKED ISSUE #285: Add --overwrite option when importing pipelines and CEs
At the moment, you get an error if you try to import a pipeline / compute environment that already exists in a Workspace. This is fine as the default behaviour, however, it would be nice to have an option to `--overwrite` or `--force` re-create any existing entities. This is particularly important when using the Tower CLI for scripting because you first have to check whether the pipeline/compute environment exists, `delete` it and then import it. Be easier for the Tower CLI to take care of this natively.

```
$ touch pipeline.json      
$ tw pipelines import --workspace=community/showcase --name=nf-core-rnaseq --compute-env=AWS_Batch_Ireland pipeline.json

 ERROR: A pipeline with name 'nf-core-rnaseq' already exists in this workspace
```

```
$ touch ce.json
$ tw compute-envs import --workspace=community/showcase --credentials=Aws_eval --name=AWS_Batch_Ireland ce.json

 ERROR: A compute environment with name 'AWS_Batch_Ireland' already exists within the workspace 'showcase'
```

There may be other `tw` commands where this could be used too. I have only ever needed this functionality when dealing with pipelines / CEs.

JaimeSeqLabs:
## Description

Adds '--overwrite' flag to pipeline and CE import commands.
If the flag is set the resource is deleted before importing the data.

Closes #285 

## Guidelines for testing
`tw compute-envs import --overwrite -n TestCE import_data.json`

`tw pipelines import --overwrite -n TestPipeline import_data.json`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -118,6 +118,33 @@ jobs:
           GITHUB_USERNAME: ${{ github.actor }}
           GITHUB_TOKEN: ${{ secrets.GH_SEQERA_TOKEN }}
           PLATFORM: osx-x86_64
+      - name: Codesign binary
+        if: "contains(github.event.head_commit.message, '[release]') && github.event.ref=='refs/heads/master'"
+        env:
+          MACOS_CERTIFICATE: ${{ secrets.MACOS_CERTIFICATE }}
+          MACOS_CERTIFICATE_PWD: ${{ secrets.MACOS_CERTIFICATE_PWD }}
+          MACOS_CERTIFICATE_NAME: ${{ secrets.MACOS_CERTIFICATE_NAME }}
+          MACOS_CI_KEYCHAIN_PWD: ${{ secrets.MACOS_CI_KEYCHAIN_PWD }}
+        run: |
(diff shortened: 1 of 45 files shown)


jordeu (approved):

tcrespog (approved):

drpatelh:
Thanks @JaimeSeqLabs ! Would it be possible to extend the scope of this PR to `--overwrite` any entity in Tower? This is particularly useful in automation scenarios where you want to re-run a script and not have to delete everything individually.

For example, if you search for `overwrite` in this file you can see it is possible to set it for any entity and not just Pipelines / CEs as requested in the linked issue.
https://github.com/seqeralabs/twkit/blob/661f4e12ecd86fbc54414a26456fbbacff4b773a/examples/yaml/twkit-e2e.yml#L7

Having this functionality in the CLI will mean we can remove the customisation we are using in `twkit`.

Thanks!

JaimeSeqLabs:
See https://github.com/seqeralabs/tower-cli/issues/338 for the aggregated list

PR #334, opened 2023-09-14
Added apple developer sign in CI

munishchouhan:
This PR will add apple developer signature steps in CI workflow for mac

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -125,6 +125,34 @@ jobs:
           name: tw_mac
           path: build/native/nativeCompile/tw
 
+      - name: Codesign binary
+        if: "contains(github.event.head_commit.message, '[release]') && github.event.ref=='refs/heads/master'"
+        env:
+          MACOS_CERTIFICATE: ${{ secrets.MACOS_CERTIFICATE }}
+          MACOS_CERTIFICATE_PWD: ${{ secrets.MACOS_CERTIFICATE_PWD }}
+          MACOS_CERTIFICATE_NAME: ${{ secrets.MACOS_CERTIFICATE_NAME }}
+          MACOS_CI_KEYCHAIN_PWD: ${{ secrets.MACOS_CI_KEYCHAIN_PWD }}
+        run: |
+          echo $MACOS_CERTIFICATE | base64 --decode > certificate.p12
+          security create-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+          security default-keychain -s build.keychain
+          security unlock-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+          security import certificate.p12 -k build.keychain -P "$MACOS_CERTIFICATE_PWD" -T /usr/bin/codesign
+          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "$MACOS_CI_KEYCHAIN_PWD" build.keychain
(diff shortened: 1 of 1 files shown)


pditommaso:
With the same certificate it can be signed more than one app right?

munishchouhan:
> With the same certificate it can be signed more than one app right?

yes
https://developer.apple.com/forums/thread/657993

.github/workflows/build.yml
```diff
@@ -125,6 +125,34 @@ jobs:
...
+          echo $MACOS_CERTIFICATE | base64 --decode > certificate.p12
+          security create-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+          security default-keychain -s build.keychain
+          security unlock-keychain -p "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+          security import certificate.p12 -k build.keychain -P "$MACOS_CERTIFICATE_PWD" -T /usr/bin/codesign
+          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "$MACOS_CI_KEYCHAIN_PWD" build.keychain
+          /usr/bin/codesign --force -s "$MACOS_CERTIFICATE_NAME" --options runtime build/native/nativeCompile/tw -v
+
+      - name: Notarize binary
+        if: "contains(github.event.head_commit.message, '[release]') && github.event.ref=='refs/heads/master'"
```
pditommaso, line 145: I think `tw` does not use `[release]` for the project release? @jordeu can you please confirm?

munishchouhan: I am using this if loop from its release job
so it should be fine 
https://github.com/seqeralabs/tower-cli/blob/d32ea6db4ee349bc9a5d9399b0053bdc6034e21c/.github/workflows/build.yml#L232

pditommaso: you are right

jordeu: Yes, it's fine.

jordeu (approved):

.github/workflows/build.yml
```diff
@@ -125,6 +125,34 @@ jobs:
...
+        env:
+          MACOS_AC_API_CERT: ${{ secrets.MACOS_AC_API_CERT }}
+          MACOS_AC_API_ISSUER_ID: ${{ secrets.MACOS_AC_API_ISSUER_ID }}
+          MACOS_AC_API_KEY_ID: ${{ secrets.MACOS_AC_API_KEY_ID }}
+        run: |
+          echo $MACOS_AC_API_CERT | base64 --decode > AuthKey.p8
+          xcrun notarytool store-credentials "notarytool-profile" -k AuthKey.p8 -d "$MACOS_AC_API_KEY_ID" -i "$MACOS_AC_API_ISSUER_ID"
+          ditto -c -k --keepParent "build/native/nativeCompile/tw" "notarization.zip"
+          xcrun notarytool submit "notarization.zip" --keychain-profile "notarytool-profile" --wait
```
pditommaso, line 154: looking more into this, it looks the codesign is applied "after" the *upload* step (see line 122), therefore I guess it just get lost

pditommaso: (same for wavelit)

munishchouhan: yes, codesign signs the same binary
and we need it to stored somewhere, before passing it to codesign

munishchouhan: ok i think i am wrong,
let me check and correct

munishchouhan: done, i will correct in wavelit too

Issue #360, opened 2023-11-09
Partial matching behavior on tw pipelines command


markpanganiban:
The pipeline subcommand will return an error when there are multiple matches with the pipeline name. 

Example below:

```
mark@LAPTOP-CURHNF2E:~/scripts/tw$ ./tw pipelines view -n 'mark-slurm' -w 'seqeralabs/testing'

 ERROR: Multiple pipelines match 'mark-slurm' at [seqeralabs / testing] workspace

mark@LAPTOP-CURHNF2E:~/scripts/tw$ ./tw pipelines view -n 'hello-mark-slurm' -w 'seqeralabs/testing'

  Pipeline at [seqeralabs / testing] workspace:

    --------------+-------------------------------------------
     ID           | 10884606264418
     Name         | hello-mark-slurm
     Description  |
     Repository   | https://github.com/markpanganiban/nf-test
     Compute env. | mark-slurm
     Labels       | No labels found
```

jordeu:
Can you give it a try to this:

```
./tw pipelines view -n '"mark-slurm"' -w 'seqeralabs/testing'
```

The option to use quotes to do a exact search was added at API level few months ago, not sure if this works at CLI.

markpanganiban:
It works with quotes!
```
$ ./tw pipelines view -n '"mark-slurm"' -w 'seqeralabs/testing'

  Pipeline at [seqeralabs / testing] workspace:

    --------------+-------------------------------------------
     ID           | 149276434439194
     Name         | mark-slurm
     Description  |
     Repository   | https://github.com/markpanganiban/nf-test
     Compute env. | mark-slurm
     Labels       | No labels found
```

jordeu:
I think that the CLI should automatically add the quotes (unless they are already added). This was not done when this was implemented because it was not supported on API side.

I don't see any use case that you want to do a partial string match here.

drpatelh:
Agree 👍🏽 Default behaviour should be exact matching.

PR #362, opened 2023-11-10
Send quote-enclosed pipeline name to API when retrieving pipeline.

robsyme:
Signed-off-by: Robert Syme <rob.syme@gmail.com>

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/pipelines/AbstractPipelinesCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/pipelines/AbstractPipelinesCmd.java
@@ -37,7 +37,7 @@ public abstract class AbstractPipelinesCmd extends AbstractApiCmd {
 
     protected PipelineDbDto pipelineByName(Long workspaceId, String name) throws ApiException {
 
-        ListPipelinesResponse list = api().listPipelines(Collections.emptyList(), workspaceId, null, null, name, "all");
+        ListPipelinesResponse list = api().listPipelines(Collections.emptyList(), workspaceId, null, null, "\"" + name + "\"", "all");
 
         if (list.getPipelines().isEmpty()) {
             throw new PipelineNotFoundException(name, workspaceRef(workspaceId));


robsyme:
Closes https://github.com/seqeralabs/tower-cli/issues/360

src/main/java/io/seqera/tower/cli/commands/pipelines/AbstractPipelinesCmd.java
```diff
@@ -37,7 +37,7 @@ public AbstractPipelinesCmd() {
 
     protected PipelineDbDto pipelineByName(Long workspaceId, String name) throws ApiException {
 
-        ListPipelinesResponse list = api().listPipelines(Collections.emptyList(), workspaceId, null, null, name, "all");
+        ListPipelinesResponse list = api().listPipelines(Collections.emptyList(), workspaceId, null, null, "\"" + name + "\"", "all");
```
jordeu, line 40: Better check that is not already quoted and escape any other quote.

Here a ChatGPT proposal:
```java
public class StringQuoter {
    public static String quoteString(String input) {
        if (input == null) {
            return "\"\"";
        }

        // Escaping intermediate quotes
        String escapedInput = input.replace("\"", "\\\"");

        // Check if the string is already quoted
        if (escapedInput.startsWith("\"") && escapedInput.endsWith("\"")) {
            return escapedInput;
        } else {
            // Adding quotes
            return "\"" + escapedInput + "\"";
        }
    }

    public static void main(String[] args) {
        // Test the function
        String testString = "Hello, \"world\"!";
        System.out.println(quoteString(testString)); // Should print: "Hello, \"world\"!"

        String alreadyQuoted = "\"Hello, world!\"";
        System.out.println(quoteString(alreadyQuoted)); // Should print the same string without changes
    }
}
```

pditommaso:
Not sure it's a good idea, why doing this?

jordeu:
> Not sure it's a good idea, why doing this?

What do you mean? my comment or the whole thing?

jordeu:
The root of the problem is that here the CLI wants to do an exact match search by pipeline name because we expect the name to be unique. And the only way of doing this at API level is enclosing the name with quotes.

jordeu:
> I see; therefore, it's the field in the request. As such, should not be managed on BE side? how the FE handle this?

The FE do not has this need because it is always first doing a list of all the pipelines in a workspace and from that list knows the internal Id of the pipeline. So FE is never querying BE for a single pipeline by name.

JaimeSeqLabs:
Closing this in favor of https://github.com/seqeralabs/tower-cli/pull/425

Issue #355, opened 2023-11-02
Add Workspace name in JSON from tw runs dump


drpatelh:
We are developing a MultiQC module that is able to scrape the output of `tw runs dump` in order to aggregate metrics into a single report across runs. It would be useful to be able to get the Workspace name directly from the JSON files that we can add to the General Stats table to clarify where the run was executed, especially if collating runs across Workspaces in the Report. 

At the moment, only the `workspaceId` is reported in `workflow-launch.json`.

See https://github.com/ewels/MultiQC/pull/2151#issuecomment-1788912050

JaimeSeqLabs:
Closed in #364

PR #364, opened 2023-11-21
Add extra metadata to workflow runs exported files

### LINKED ISSUE #356: Add Run URL in JSON from tw runs dump
We are developing a MultiQC module that is able to scrape the output of `tw runs dump` in order to aggregate metrics into a single report across runs. It would be useful to be able to get the Run URL directly from the JSON files that we can add to the report. This will allow us to add links to the Report that are able to open the Run directly on the Seqera Platform.

See https://github.com/ewels/MultiQC/pull/2151#issuecomment-1788912050

### LINKED ISSUE #357: Add Run labels in JSON from tw runs dump
We are developing a MultiQC module that is able to scrape the output of `tw runs dump` in order to aggregate metrics into a single report across runs. It would be useful to be able to get run labels directly from the JSON files that we can add to the report. We can then add these to the report to help with filtering / sorting etc.

See https://github.com/ewels/MultiQC/pull/2151#issuecomment-1788912050

### LINKED ISSUE #361: Add keys to `tw runs dump` output to complete metadata for a workflow
Is it possible to add the following keys in the `workflow.json` output generated by `tw runs dump`? This would provide a complete set of metadata for a given run/workflow and has been requested by users intending to incorporate this metadata into sources downstream of the platform (i.e. databases, dashboards).

- [ ] `pipelineId` (unique identifier for pipelines on the Launchpad, obtained when running `tw pipelines list`)
- [ ] `workspaceId`
- [ ] `userEmail` (we include `userName` currently)
- [ ] `userId`

JaimeSeqLabs:
## Description

Closes  #361 #357 #356 #355 

Introduces a new export file specific to workflow metadata and aggregated data that does not belong to `Workflow` entity. 

This new file `workflow-metadata.json` generated by `tw runs dump` command containing the following fields:
+ `pipelineId`
+ `workspaceId`
+ `workspaceName`
+ `userId`
+ `runUrl`
+ `labels`

## Guidelines for testing
1. Launch a pipeline
2. Copy the workflow run ID
3. Run `tw runs dump -w <your_workspace> -i <workflow_run_id> -o test-runs-dump.tar.gz`
4. The expected console output is:
```
- Tower info
- Workflow details
- Task details

  Pipeline run '2sk2IPkODYQqUY' at [Org / Wsp] workspace details dump at 'test-runs-dump.tar.gz'

```
5. Inside the tar file, the `workflow-metadata.json` file should contain the following workflow metadata fields:
```
{
  "pipelineId": ...,
  "workspaceId": ...,
  "workspaceName": ...,
  "userId": ...,
  "runUrl": ...,
  "labels": [
    ...
  ]
}
```


src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
```diff
@@ -167,15 +170,36 @@ private void dumpTasks(PrintWriter progress, TarArchiveOutputStream out, Long ws
     private void dumpWorkflowDetails(PrintWriter progress, TarArchiveOutputStream out, Long wspId) throws ApiException, IOException {
         progress.println(ansi("- Workflow details"));
 
+        // User info
+        UserDbDto user = api().userInfo().getUser();
```
jordeu, line 174: This is the user that is doing the dump. But maybe, at the issue, @ejseqera refers to the user that launched the workflow.

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
```diff
@@ -167,15 +170,36 @@ private void dumpTasks(PrintWriter progress, TarArchiveOutputStream out, Long ws
...
+            }
+        }
+
+        // Load and metrics info
         WorkflowLoad workflowLoad = workflowLoadByWorkflowId(wspId, id);
-        Launch launch = workflow.getLaunchId() != null ? launchById(wspId, workflow.getLaunchId()) : null;
         List<WorkflowMetrics> metrics = api().describeWorkflowMetrics(workflow.getId(), wspId).getMetrics();
-        addEntry(out, "workflow.json", Workflow.class, workflow);
+
+        WorkflowDumpExportFormat wfDetails = new WorkflowDumpExportFormat(workflow, pipelineId, wspId, user.getId(), user.getEmail());
```
jordeu, line 200: For me to create a new class that extends `Workflow` class with it's special serializer it's a bit dangerous. At some point `Workflow` can add new attributes that clash with these attributes.
 
If it's a real requirement to add the attributes in the same `workflow.json` then seems ok to me. But most likely to put all this data into a new JSON (example: `workflow-info.json`) is also a valid solution for our users and the implementation is easier and a bit more safe. 

What do you think @ejseqera ?

src/main/java/io/seqera/tower/cli/shared/WorkflowDumpExportFormat.java
```diff
@@ -0,0 +1,190 @@
...
+    public static abstract class WorkflowDumpMixin {}
+
+    public static String serialize(final WorkflowDumpExportFormat wfExport) throws JsonProcessingException {
+        return buildMapper()
+                .writerWithDefaultPrettyPrinter()
+                .withAttribute("pipelineId", wfExport.getPipelineId())
+                .withAttribute("workspaceId", wfExport.getWorkspaceId())
+                .withAttribute("userId", wfExport.getUserId())
+                .withAttribute("userEmail", wfExport.getUserEmail())
+                .writeValueAsString(wfExport.getWorkflow());
```
endre-seqera, line 118: Thought: related to @jordeu 's point above: What happens if in the future `Workflow` class will have attributes with the same name as the envelope class `WorkflowDumpExportFormat`, like `pipelineId` present in both?

Probably both `pipelineId` attributes will mean the same thing, they won't be different, so that might not be an issue. 

But I'm curios what will be the behavior of the serializer. Will it throw an error - duplicate name in JSON? Or will it overwrite, the later write (for `Workflow`) will overwrite the former write (for `wfExport`)? It could be helpful to have a unit test verifying this behavior.

src/main/java/io/seqera/tower/cli/shared/WorkflowDumpExportFormat.java
```diff
@@ -0,0 +1,190 @@
...
+    private Long pipelineId;
+
+    private Long workspaceId;
+
+    private Long userId;
+
+    private String userEmail;
+
+    public WorkflowDumpExportFormat(
+            final Workflow workflow,
```
pgeadas, line 57: I think that it is a bit weird that the parameters being passed are `final` but then the fields of the instance are not. Any reason behind this?

src/test/java/io/seqera/tower/cli/runs/RunsCmdTest.java
```diff
@@ -553,4 +554,65 @@ void testDownloadTaskLog(MockServerClient mock) throws IOException {
         assertEquals(new RunFileDownloaded(file, RunDownloadFileType.stdout).toString(), out.stdOut);
         assertEquals(0, out.exitCode);
     }
+
+    @Test
+    void testDumpRuns(MockServerClient mock) throws IOException {
+
+        mock.when(
```
pgeadas, line 561: Same for the test, I think it would be nice to abstract the duplicated code into a single method.

src/test/java/io/seqera/tower/cli/shared/WorkflowDumpExportFormatTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+import static io.seqera.tower.cli.utils.JsonHelper.parseJson;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class WorkflowDumpExportFormatTest {
+
+    @Test
+    void testFormatDeserialization() throws JsonProcessingException {
+        String json = "{\n" +
```
pgeadas, line 32: I think that this pattern of hardcoding huge chunks of text (in this case even worse since it is Json and we need to escape the ") should be avoided. We could add the json files to the resources folder (maybe using the test name as a folder name, to separate the resources from other tests) and load them. That way I think the class becomes more readable and less error prone. What do you think?

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
ejseqera: Not necessarily a real requirement to add the attribute in `workflow.json`, it can be elsewhere, as long as it is captured somewhere. Pulling metadata that pertains to the who/what/when/how into `workflow-info` wouldn't be bad at all.

Will pulling it in `workflow-info.json` as an example, make it easier to retrieve additional attributes in the future that we may not be capturing now?

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
ejseqera: Yes, to clarify, we are looking for info of the user that launched the workflow.

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
JaimeSeqLabs: > Will pulling it in workflow-info.json as an example, make it easier to retrieve additional attributes in the future that we may not be capturing now?

Yes, that would make it a lot easier because that way we don't need to artificially add extra fields to the base entity (`Workflow` entity), both files can evolve separately without fear of overwriting each other fields

I will add the separate metadata file then

src/main/java/io/seqera/tower/cli/shared/WorkflowDumpExportFormat.java
JaimeSeqLabs: Yeah, this is risky but unfortunately there is no other way of adding "virtual" fields to JSON.
We are going to separate the added metadata into another file so hopefully that will make this future-proof.

src/main/java/io/seqera/tower/cli/shared/WorkflowDumpExportFormat.java
JaimeSeqLabs: A mistake, `final` is missing from the fields.
I'll fix it

src/main/java/io/seqera/tower/cli/shared/WorkflowDumpExportFormat.java
```diff
@@ -0,0 +1,190 @@
...
+        public WorkflowDumpExportFormatDeserializer() {
+            this(null);
+        }
+
+        public WorkflowDumpExportFormatDeserializer(Class<WorkflowDumpExportFormat> vc) {
+            super(vc);
+        }
+
+        @Override
+        public WorkflowDumpExportFormat deserialize(JsonParser parser, DeserializationContext ctx) throws IOException, JsonProcessingException {
```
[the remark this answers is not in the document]
JaimeSeqLabs: Yes this looks cleaner

src/test/java/io/seqera/tower/cli/runs/RunsCmdTest.java
JaimeSeqLabs: Do you mean setting up the test endpoints in a separate method?

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
```diff
@@ -167,15 +171,47 @@ private void dumpTasks(PrintWriter progress, TarArchiveOutputStream out, Long ws
...
         DescribeWorkflowResponse workflowResponse = workflowById(wspId, id);
         Workflow workflow = workflowResponse.getWorkflow();
         if (workflow == null) {
             throw new TowerException("Unknown workflow");
         }
+
+        // Launch info
+        Launch launch = null;
+        Long pipelineId = null;
+        if (workflow.getLaunchId() != null) {
```
pgeadas, line 184: Can't we create a separate method with a descriptive name of what is happening here, instead of directly put the if clause?

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
```diff
@@ -167,15 +171,47 @@ private void dumpTasks(PrintWriter progress, TarArchiveOutputStream out, Long ws
...
+
+            launch = launchById(wspId, workflow.getLaunchId());
+
+            DescribeWorkflowLaunchResponse wfLaunchResponse = workflowLaunchById(wspId, workflow.getId());
+            if (wfLaunchResponse != null && wfLaunchResponse.getLaunch() != null) {
```
pgeadas, line 189: Same here. What is the "if" checking? I think it would be better to create a separate method that returns a boolean, and has a descriptive name of what we are actually checking. Example: `isLaunchResponseValid()` or something.

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
```diff
@@ -167,15 +171,47 @@ private void dumpTasks(PrintWriter progress, TarArchiveOutputStream out, Long ws
...
+                pipelineId = wfLaunchResponse.getLaunch().getPipelineId();
+            }
+        }
+
+        // User info
+        Long userId = workflow.getOwnerId();
+        String userMail = null;
+        String userName = workflow.getUserName();
+        try {
```
pgeadas, line 198: Same here. This Try catch could be hidden on a separate method with a descriptive name. That way all this method becomes way simpler and easy to understand.

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
JaimeSeqLabs: I think the `if` condition is clear. We check if we got a response and if the response contains a `Launch` object (we don't have the null chain operator in Java).
Creating a separate method just for this simple check is not justified in my opinion, although It would be if the condition check is re-used multiple times in the class.

src/main/java/io/seqera/tower/cli/commands/runs/DumpCmd.java
JaimeSeqLabs: I think that moving fragments of the code to a separate method makes sense if said method is getting reused more than once or if the fragment is overly complex. Otherwise we are just scattering the code around without actually reducing complexity.
I'll add some comments given that this filtering fragment is a bit convoluted. If we find an easier approach I'll update it.

JaimeSeqLabs: Comments are not pollution in the code base. We strongly encourage adding comments wherever is necessary as long as they explain the "why" rather than the "what" in the code.

> Which one looks cleaner?

This is subjective, sure the wrapper method is 3 lines but then you have moved the same code into two methods, so you end with N+3 lines and additional jumps that makes debugging more difficult.

P.S: Thanks a lot for the recommendation, I enjoyed reading that book back in the day. Let me return the favor and recommend you another [book](https://www.amazon.com/Pragmatic-Programmer-Journeyman-Master/dp/020161622X).
If you are more curious about this and other pragmatic principles please let me know.

pgeadas: > moved the same code into two methods

Not really, what happens is that both methods will call a single method where all the duplicate code is. However, that part is not shown in the screenshot.

> makes debugging more difficult
How so?

Thanks for the book recommendation 👍🏻

JaimeSeqLabs:
I linked all the three issues related to workflow metadata in this PR, their implementations are similar one to another.

drpatelh:
Thank you for this guys! 🙏🏽 

With regard to this issue https://github.com/seqeralabs/tower-cli/issues/357

Would we be able to add pipeline labels to the default JSON file created by `tw -o runs view` too please? 

At the moment, resource labels are dumped but we need the pipeline labels to filter runs for our QuickLaunch implementations:

```
~ tw -o json runs view --workspace=community/showcase --id=1v9SlFbY4LR7wl       
{
  "general" : {
    "id" : "1v9SlFbY4LR7wl",
    "operationId" : "f2c38f50-a801-461c-9527-bb4be471359e",
    "runName" : "ridiculous_ritchie",
    "startingDate" : "2023-11-07T16:53:43Z",
    "commitId" : "3bec2331cac2b5ff88a1dc71a21fab6529b57a0f",
    "sessionId" : "e6b26621-e067-47a4-9523-ff46b98fb74f",
    "username" : "jonathanyoung305",
    "workdir" : "s3://nf-tower-bucket/scratch/1v9SlFbY4LR7wl",
    "container" : "",
    "executors" : "awsbatch",
    "computeEnv" : "AWS_Batch_Ireland_FusionV2_NVMe",
    "nextflowVersion" : "23.08.0-edge",
    "status" : "SUCCEEDED",
    "labels" : "owner=seqera,rnaseq,workspace=showcase"
  }
}
```

drpatelh:
Actually scratch the request in my former comment. Looks like you can do this already. I tested with a pipeline that didn't have pipeline labels 🤦🏽 

```
$ tw -o json runs view --workspace=seqeralabs/showcase --id=1bGQVwwOs1zHt6
{
  "general" : {
    "id" : "1bGQVwwOs1zHt6",
    "operationId" : "10e64152-4799-46a3-bd3c-e2e45a6aac28",
    "runName" : "intergalactic_torvalds",
    "startingDate" : "2023-11-28T17:17:52Z",
    "commitId" : "94bd58250591c29e23d9afbb9deedf95800fe3fc",
    "sessionId" : "c1499562-9513-4d67-b3fe-0e86777603aa",
    "username" : "drpatelhh",
    "workdir" : "s3://seqeralabs-showcase/scratch/1bGQVwwOs1zHt6",
    "executors" : "awsbatch",
    "computeEnv" : "seqera_aws_ireland_fusionv2_nvme",
    "nextflowVersion" : "23.11.0-edge",
    "status" : "SUCCEEDED",
    "labels" : "gpu,owner=harshil,structure-prediction,workspace=showcase"
  }
}
```

Would be nice however, if we can disambiguate between pipeline labels and resource labels rather than having everything in one `labels` field. But that can come as a separate feature request.

JaimeSeqLabs:
> Would be nice however, if we can disambiguate between pipeline labels and resource labels rather than having everything in one `labels` field. But that can come as a separate feature request.

If we really needed it can be separated into a feature, sure.
But is it justified? resource vs non-resource can be distinguished by filtering the labels containing '='.

JaimeSeqLabs:
@ejseqera , @drpatelh I'm trying to refactor the method to obtain the launch user email because the current approach feels very hacky.

We are taking the userEmail from the workspace participant list by matching the userName from the workflow to the participant userName.

This approach doesn't work if we dump the runs in the user context (there is no workspace nor participants) so in that case there is no userMail field in the json file.

Are you extracting the userMail field with a different technique? Otherwise I'll need to make changes on Seqera Platform side to include the mail in the wokflow response.

JaimeSeqLabs:
Hi @ejseqera @drpatelh ,
After some discussions in the backend team we concluded that there is no safe way of exposing the launcher user email without extensive modifications or risking personal data (email) leaks.
The closest that we can get to the requested functionality is the workspace participant scan which doesn't work for user contexts.

Sadly we have to de-scope the email metadata.

PR #368, opened 2023-11-23
notask: fixed tw script to support zsh and bash (tested on mac)

pgeadas:
## Description

Fix the tw script to detect which shell we are on, because the syntax changes a bit from bash

## Testing guidelines

1. Check if the script still works from bash
2. Test from linux (currently tested from Mac and Zsh)

### DIFF
--- a/tw
+++ b/tw
@@ -1,6 +1,14 @@
 #!/bin/bash
 
+current_shell=$(basename -- "$SHELL")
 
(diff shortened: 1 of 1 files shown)


jordeu (approved):
working with Linux + Bash

JaimeSeqLabs (approved):

Issue #394, opened 2024-03-26
Cannot import compute environment from file when Labels are included


k1sauce:
As the title says when I am trying to make a compute environment via

`tw compute-envs import -n test1 test-no-label.json`

everything is ok
`New AWS-BATCH compute environment 'test1' added at [ClearNote_Health / discovery-pipeline-prod] workspace
`
however when I am trying to use a pre-existing resource label

`tw compute-envs import -n test1 test.json`
  
I get the following error
 
```
java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to io.seqera.tower.model.LabelDbDto
	at java.base@17.0.8/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base@17.0.8/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base@17.0.8/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base@17.0.8/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base@17.0.8/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base@17.0.8/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base@17.0.8/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at io.seqera.tower.cli.commands.computeenvs.ImportCmd.exec(ImportCmd.java:68)
	at io.seqera.tower.cli.commands.AbstractApiCmd.call(AbstractApiCmd.java:413)
	at io.seqera.tower.cli.commands.AbstractApiCmd.call(AbstractApiCmd.java:65)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1953)
	at picocli.CommandLine.access$1300(CommandLine.java:145)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2358)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2352)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2314)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2179)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2316)
[...]

PR #396, opened 2024-03-27
fix(labels-import): pass the type info to the mapper

k1sauce:
https://github.com/seqeralabs/tower-cli/issues/394 

Fixes the issue when creating a compute environemnt from a json file with the labels field


k1sauce:
I suppose that there is a better way to make extractJsonNodeValue generic, though as far as I can tell it is only be used to parse the labels node

JaimeSeqLabs (approved):

PR #424, opened 2024-06-18
Launch with optimization turned off

### LINKED ISSUE #393: Enable or disable pipeline optimization via CLI
If the pipeline optimization (lightbulb) is enabled, there is no way to enable or disable when launching from the CLI. If the optimization is inappropriate, this causes the pipeline to fail with no way of bypassing it.

I would like to be able to toggle optimization on or off at launch time.

JaimeSeqLabs:
## Description

Closes #393 

This PR adds support for disabling the optimized configuration when launching a pipeline by setting the `--disable-optimization` flag.

## Guidelines for testing

1. Create a new pipeline inside an organization workspace (optimization is available in non-personal workspaces only).
2. Launch the pipeline once so the optimization becomes available.
image
3. Click on the light bulb and save the default optimization targets and values. 
image
4. Verify that the pipeline is shown as optimized.
image
5. Launch the optimized pipeline and check that the configuration contains the optimized parameters.
image
6. Launch the optimized pipeline using `./tw` with optimization turned off and check that the configuration does **NOT** contain the optimized parameters.
```bash
./tw launch HelloOptimization -w JaimeInc/Wsp --name=ThisShouldNotHaveOptimizations --disable-optimization
```
image


pgeadas:
Do I need to do anything special to get the optimization available? 
image

I created a pipeline inside a workspace and ran it, but still can't see the lamp.

JaimeSeqLabs:
I'm assuming you are running the pipeline in Seqera Platform (not in a local dev env)
Did the pipeline finish successfully?

>I created a pipeline inside a workspace and ran it, but still can't see the lamp.
Is the lamp icon shown as unavailable or is it not shown at all?

pgeadas:
> I'm assuming you are running the pipeline in Seqera Platform (not in a local dev env)

It can't be run locally?

JaimeSeqLabs:
> It can't be ran locally? 

It can, of course. But, given that tower-cli is a public repo, the test should be expected to run against prod so anyone can try it.

We can discuss offline the setup for the local instance.

swampie:
tested! works fine!

swampie (approved):

PR #447, opened 2024-08-30
Update LaunchOptions.java

llewellyn-sl:
Improve --config flag description (overrides default Nextflow conf, not additive)

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/pipelines/LaunchOptions.java
+++ b/src/main/java/io/seqera/tower/cli/commands/pipelines/LaunchOptions.java
@@ -39,7 +39,7 @@ public class LaunchOptions {
     @Option(names = {"--revision"}, description = "A valid repository commit Id, tag or branch name.")
     public String revision;
 
-    @Option(names = {"--config"}, description = "Path to a Nextflow config file. Values defined here override the same values in the pipeline repository config file.")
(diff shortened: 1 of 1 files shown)


src/main/java/io/seqera/tower/cli/commands/pipelines/LaunchOptions.java
```diff
@@ -39,7 +39,7 @@ public class LaunchOptions {
     @Option(names = {"--revision"}, description = "A valid repository commit Id, tag or branch name.")
     public String revision;
 
-    @Option(names = {"--config"}, description = "Additional Nextflow config file.")
+    @Option(names = {"--config"}, description = "Path to a Nextflow config file. Values defined here override the same values in the pipeline repository config file.")
```
mbosio85, line 42: I think it's not precise, or at least it seems to me that the override is key-based while it's not the case.
It is a total override, discarding completely what comes from the pipeline.

llewellyn-sl, line 42:
```suggestion
    @Option(names = {"--config"}, description = "Path to a Nextflow config file. Values defined here override the same values in the pipeline repository config file, and all configuration values specified in Platform pipeline or compute environment Nextflow config fields are ignored.")
```

mbosio85 (approved):

PR #477, opened 2025-01-23
PLAT-1242 - Add DataStudio start command.

georgi-seqera:
### **Context**

Work relates to task: https://seqera.atlassian.net/browse/PLAT-1242

Summary: add the ability to start a data studio via the CLI.
Optionality to:
* override/update any of the existing DataStudio configurations (gpu, cpu, mountData, etc.) or description
* have command wait for specific status of DataStudio

### **Testing**

**Trigger a start (default asyncrounous command):**
```
$ ./tw studios start -w 27230932650799 -i 6c74268d

  Data Studio 6c74268d START successfully submitted at [data-studios / data-studios] workspace.

     https://dev-tower.net/orgs/data-studios/workspaces/data-studios/studios/6c74268d/connect
```
https://github.com/user-attachments/assets/7226c2ff-7e16-4645-a81e-5243ffd66a37

**Trigger a start with override configuration supplied**
```
$ ./tw studios start -w 27230932650799 -i 6c74268d --description 'Override test description' --cpu 4

  Data Studio 6c74268d START successfully submitted at [data-studios / data-studios] workspace.

     https://dev-tower.net/orgs/data-studios/workspaces/data-studios/studios/6c74268d/connect
```
https://github.com/user-attachments/assets/4e11c6ad-5ae5-4b46-93b4-51b9bfa5768d

**Trigger a start with `--wait` flag to wait until desired DataStudio status is reached.**
```
$ ./tw studios start -w 27230932650799 -i 6c74268d --wait RUNNING

  Data Studio 6c74268d START successfully submitted at [data-studios / data-studios] workspace.

     https://dev-tower.net/orgs/data-studios/workspaces/data-studios/studios/6c74268d/connect

  Waiting running status....
  Status transitioned to: starting
  Provisioning compute resources...................
  Mounted checkpoint 8 of 9
  Status transitioned to: running  [DONE]
```
https://github.com/user-attachments/assets/2893690b-2f1c-4d44-b79b-5f1f32edc38f

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/datastudios/DataStudioConfigurationOptions.java
+++ b/src/main/java/io/seqera/tower/cli/commands/datastudios/DataStudioConfigurationOptions.java
@@ -35,7 +35,4 @@ public class DataStudioConfigurationOptions {
     @CommandLine.Option(names = {"--mount-data"}, description = "Optional configuration override for 'mountData' setting (comma separate list of datalinkIds)", split = ",")
     public List<String> mountData;
 
-    @CommandLine.Option(names = {"--conda-env"}, description = "Optional configuration override for 'condaEnvironment' setting (YAML conda packages configurations)")
-    public String condaEnvironment;
-
 }
--- a/src/main/java/io/seqera/tower/cli/commands/datastudios/DataStudioConfigurationOptions.java
+++ b/src/main/java/io/seqera/tower/cli/commands/datastudios/DataStudioConfigurationOptions.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2021-2023, Seqera.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.seqera.tower.cli.commands.datastudios;
+
+import java.util.List;
+
+import picocli.CommandLine;
+
+public class DataStudioConfigurationOptions {
+
+    @CommandLine.Option(names = {"-g", "--gpu"}, description = "Optional configuration override for 'gpu' setting (Integer representing number of cores)")
+    public Integer gpu;
+
+    @CommandLine.Option(names = {"-c", "--cpu"}, description = "Optional configuration override for 'cpu' setting (Integer representing number of cores)")
+    public Integer cpu;
+
+    @CommandLine.Option(names = {"-m", "--memory"}, description = "Optional configuration override for 'memory' setting (Integer representing memory in MBs)")
+    public Integer memory;
+
+    @CommandLine.Option(names = {"--mount-data"}, description = "Optional configuration override for 'mountData' setting (comma separate list of datalinkIds)", split = ",")
+    public List<String> mountData;
+
+    @CommandLine.Option(names = {"--conda-env"}, description = "Optional configuration override for 'condaEnvironment' setting (YAML conda packages configurations)")
+    public String condaEnvironment;
+
+}
--- a/src/main/java/io/seqera/tower/cli/utils/ResponseHelper.java
+++ b/src/main/java/io/seqera/tower/cli/utils/ResponseHelper.java
@@ -173,10 +173,7 @@ public class ResponseHelper {
                     lastReported = status;
                 }
                 if (additionalProgressSteps != null) {
-                    String progressStep = additionalProgressSteps.get();
-                    if (progressStep != null) {
-                        out.print(progressStep);
-                    }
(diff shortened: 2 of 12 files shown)


weronikasosnowskaseqera (commented):
I think the address we should return is different. The one you're returning now is used by FE to load the iframe. If you go to UI and click `Copy data studio URL` then you will get this URL (example) http://tower.localhost:8000/orgs/org/workspaces/wsp/studios/d3bf39a8/connect

src/main/java/io/seqera/tower/cli/commands/datastudios/DataStudioConfigurationOptions.java
```diff
@@ -0,0 +1,41 @@
...
+    @CommandLine.Option(names = {"-c", "--cpu"}, description = "Optional configuration override for 'cpu' setting (Integer representing number of cores)")
+    public Integer cpu;
+
+    @CommandLine.Option(names = {"-m", "--memory"}, description = "Optional configuration override for 'memory' setting (Integer representing memory in MBs)")
+    public Integer memory;
+
+    @CommandLine.Option(names = {"--mount-data"}, description = "Optional configuration override for 'mountData' setting (comma separate list of datalinkIds)", split = ",")
+    public List<String> mountData;
+
+    @CommandLine.Option(names = {"--conda-env"}, description = "Optional configuration override for 'condaEnvironment' setting (YAML conda packages configurations)")
```
endre-seqera, line 38: issue: when starting an already created DataStudio conda-env cannot be overwritten, this command line option should be removed

src/main/java/io/seqera/tower/cli/commands/datastudios/DataStudioConfigurationOptions.java
```diff
@@ -0,0 +1,41 @@
...
+    @CommandLine.Option(names = {"-g", "--gpu"}, description = "Optional configuration override for 'gpu' setting (Integer representing number of cores)")
+    public Integer gpu;
+
+    @CommandLine.Option(names = {"-c", "--cpu"}, description = "Optional configuration override for 'cpu' setting (Integer representing number of cores)")
+    public Integer cpu;
+
+    @CommandLine.Option(names = {"-m", "--memory"}, description = "Optional configuration override for 'memory' setting (Integer representing memory in MBs)")
+    public Integer memory;
+
+    @CommandLine.Option(names = {"--mount-data"}, description = "Optional configuration override for 'mountData' setting (comma separate list of datalinkIds)", split = ",")
```
endre-seqera, line 35: thought:
While this works, it's a bit uncomfortable to use, since the user needs to provide data link IDs for the mounts (`--mount-data v1-cloud-YnVja2V0OndzOjI3MjMwOTMyNjUwNzk5OnMzOi8vYWRyaWFuLW5hdmFycm8tdGVzdA==`). 

When using the SeqeraPlatform, through the browser, these data link IDs are nicely hidden, and never exposed to the user (since we just need them internally). 

The ideal situation would be to achieve the same with the cli, user being able to mount a data link by reference path (`s3://adrian-navarro-test`) or name (`my-custom-data-link`).

This is just a thought to improve this in a follow-up task. I'm facing this very same issue for the `add` command, on how to specify mounted data link configuration.

src/main/java/io/seqera/tower/cli/utils/ResponseHelper.java
```diff
@@ -173,6 +186,12 @@ public static <S extends Enum<?>> Integer waitStatus(PrintWriter out, boolean sh
         } while (currentPos < targetPos && !immutableStates.contains(status));
 
         if (showProgress) {
+            if (additionalProgressSteps != null) {
+                String progressStep = additionalProgressSteps.get();
+                if (progressStep != null) {
+                    out.print(progressStep);
+                }
```
endre-seqera, line 193: idea: could this block - here and above - maybe be simplified to `out.print(additionalProgressSteps.get())` , by and making the `get()` return empty string instead of null?

endre-seqera (commented):
Nice work, only small comments!

src/main/java/io/seqera/tower/cli/commands/datastudios/DataStudioConfigurationOptions.java
georgi-seqera: Yeaah, I had the same thought and was thinking that this can probably be even taken further to have the studio itself be passed in by name, rather than id? or have the option for either id, or studioName supplied (and behind the scenes the cli will query to find the right Ids)

Then it would be more similar to how the command line works for the pipeline launch where it's just `tw launch my-pipeline` ?

endre-seqera: Start studio by name, that's a great idea! It's up to you to decide if you want to include it in this PR or not

t0randr: I think at least in few other places, we support both by name and by id.

t0randr:
Nice work.

src/main/java/io/seqera/tower/cli/commands/datastudios/DataStudioConfigurationOptions.java
georgi-seqera: I think I will to extend that as part of another PR and also take a stab at doing the same for the mountData.

endre-seqera (approved):

PR #490, opened 2025-02-12
Make autoStart option consistent with the other options

mribeirodantas:
There are options following the `--mount-data` standard, and then `--autoStart`. This PR makes the latter consistent with the other options (`--auto-start`).


georgi-seqera (approved):
Thank you very much 🫡

endre-seqera (approved):

PR #495, opened 2025-03-06
Update version to 0.11.1 for new [release]

georgi-seqera:
Creating new release with new options for session management of Studios. [PLAT-1326](https://seqera.atlassian.net/browse/PLAT-1326)

#### Changes to existing studio commands/options:
* `start-as-new` command renamed to `add-as-new`
* the autostart option has been renamed for consistency from `--autoStart` to `--auto-start`
* for mount-data option has been renamed for clarity from `--mount-data-refs` to `--mount-data-uris` for when providing a data-link URI (e.g. s3://nextflow-bucket,s3://another-bucket)

Note: @ejseqera based on feedback from people at demos, we have done some renaming mentioned above of options like the `--autoStart` and `--mount-data-refs` that I think would require a corresponding change in [seqerakit](https://github.com/seqeralabs/seqera-kit/blob/870c4f7689ebc7fd7ca4795dbf39481ed36e4371/templates/studios.yml#L19-L20).

Since the studios functionality is pretty new, we've opted to just rename it directly, instead of deprecating and creating separate one. Would it be an easy fix to rename these in seqerakit?

#### New studio functionality/options:
* Add `--lifespan` option to studios add, add-as-new and start commands.
* Add `--private` option flag to studios add, add-as-new command
* Add `--labels` option taking key=value pairs list e.g. `--labels env=dev,compute=aws`
   * Currently, the labels option will set the labels as they are (not appending) - this is consistent with the other commands in the CLI
   * Labels are searched to find existing ones and if not found, are created (creation permissioning relies on the backend endpoint role authorisation)
   * for `add` command, if `--labels` not provided - backend defaults to compute-env labels
   * for `add-as-new` command, if --labels not provided - backend defaults to copy over labels from parent studio
   * labels can be removed by specifying `--labels=""`
   * Add `--labels` option flag to studios list command to enrich labels to studios list output

[PLAT-1326]: https://seqera.atlassian.net/browse/PLAT-1326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


endre-seqera (approved):

tcrespog (commented):
Just a couple of comments for the record, as the latest version of the CLI cannot be used yet due to the Cloud API version being `1.30.0` and the minimum CLI version being `1.31.0`. No big deal, it should catch up in the next release.

PR #496, opened 2025-03-11
Propagate CE pre-run and post-run scripts on pipeline launch

tcrespog:
### Description

- Replicate behaviour found in the GUI, which sets the value of the pre-run and post-run scripts to those of the CE if they are undefined.
- Consider empty strings to be undefined values (equivalent to `null`).

### Guidelines for testing

- [ ] Using the CLI, launch a pipeline **with** pre-run script, post-run script and Nextflow config defined: the values of the pipeline are the ones that are taken into account.
- [ ] Using the CLI, launch a pipeline **with** pre-run script, post-run script and Nextflow config defined using a CE that also has pre-run script, post-run script and Nextflow config defined: the values of the pipeline are the ones that are taken into account.
```
export TOWER_API_ENDPOINT=...
export TOWER_WORKSPACE_ID=...
export TOWER_ACCESS_TOKEN=...

tw launch --compute-env my-compute-env my-pipeline
```
- [ ] Using the CLI, launch a pipeline **without** pre-run script, post-run script and Nextflow config defined using a CE that also has pre-run script, post-run script and Nextflow config defined: the  values of the CE are the ones that are taken into account.


daria-seqera:
Hm, I've been trying to test this behaviour locally, however it works for me even before this fix (tried both the natively-compiled version from `master` and the linux one from the latest release).
What I'm doing is to run the following command:
```
TOWER_API_ENDPOINT=http://localhost:8080 ./tw --insecure launch test-cli-ce-pre-post --post-run=../pre-run-sh
```
`test-cli-ce-pre-post` has both scripts defined in the CE, and I can see that the `post-run` is being overwritten by my argument

Also, I wasn't sure whether you meant that the scripts are defined when you add a pipeline, or when you launch it, so I tested the other scenario too, where I have the scripts defined in the CE, and then overwritten when adding the pipeline. Not providing parameters from the cli:
```
TOWER_API_ENDPOINT=http://localhost:8080 ./tw --insecure launch test-ce-pre-post-redefined
```
and it also worked correctly.


What am I missing?

weronikasosnowskaseqera (approved):
Hey @tcrespog do we want to move this forward?

JaimeSeqLabs (approved):

tcrespog:
> Hey @tcrespog do we want to move this forward?

Hi, @weronikasosnowskaseqera. I realized that the `nextflowConfig` field has the same behaviour as `preRunScript` and `postRunScript`; however, `nextflowConfig` is not present among the class fields, so it requires an explicit update of the SDK. I was planning to do that, but had to address other matters.

georgi-seqera:
Hey @tcrespog - I have a [PR](https://github.com/seqeralabs/tower-cli/pull/514) with another version bump of the tower-java-sdk to a version that would supersede the one in this PR. 
Would that be safe to merge without this PR or is this one a prerequisite to be merged in first? 
The changes in this one look backwards compatible to me, but double checking.

PR #505, opened 2025-04-16
decrease api version, add docs

endre-seqera:
Decrease needed API version, as it was bumped too far ahead.
Add documentation to VERSION-API to avoid possible future occurrence


pgeadas (approved):

VERSION-API
```diff
@@ -1 +1,3 @@
-1.38.0
+1.37.0
+// Only first line of this file is read
+// This version should never be bumped higher, then the current Platform API Version deployed in Cloud Production: https://cloud.seqera.io/api/service-info
```
tcrespog, line 3: What do you think about also including something such as "the version should be bumped to the minimum version where dependent API changes were introduced"? I'm saying this because, even if a particular version of the API/SDK is used, the functionalities implemented with it might have been available since previous versions: this is what happened [here](https://github.com/seqeralabs/tower-cli/pull/495#discussion_r1991794603), for example.

tcrespog (approved):

georgi-seqera (approved):

Issue #504, opened 2025-04-15
Runtime reflection error when using JSON output format


ejseqera:
When trying to list organizations or workspaces with JSON output format (`tw -o json organizations list`), the command fails with a GraalVM native-image reflection registration error.

The command fails with the following error:
```
Exception in thread "main" org.graalvm.nativeimage.MissingReflectionRegistrationError: The program tried to reflectively invoke method public java.lang.String io.seqera.tower.model.OrgType.getValue() without it being registered for runtime reflection. Add public java.lang.String io.seqera.tower.model.OrgType.getValue() to the reflection metadata to solve this problem. See https://www.graalvm.org/latest/reference-manual/native-image/metadata/#reflection for help.
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.reflect.MissingReflectionRegistrationUtils.forQueriedOnlyExecutable(MissingReflectionRegistrationUtils.java:83)
        at java.base@21.0.6/java.lang.reflect.Method.acquireMethodAccessor(Method.java:77)
        at java.base@21.0.6/java.lang.reflect.Method.invoke(Method.java:577)
        at com.fasterxml.jackson.databind.introspect.AnnotatedMethod.getValue(AnnotatedMethod.java:183)
        at com.fasterxml.jackson.databind.ser.std.JsonValueSerializer.serialize(JsonValueSerializer.java:166)
        at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:727)
        at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:722)
        at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:166)
        at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serializeContents(IndexedListSerializer.java:119)
        at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serialize(IndexedListSerializer.java:79)
        at com.fasterxml.jackson.databind.ser.impl.IndexedListSerializer.serialize(IndexedListSerializer.java:18)
        at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:727)
        at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:722)
[...]

PR #506, opened 2025-04-16
fix: add missing method in reflect-config.json for OrgType enum

georgi-seqera:
Should fix issue: https://github.com/seqeralabs/tower-cli/issues/504


endre-seqera:
Included in https://github.com/seqeralabs/tower-cli/pull/505

PR #513, opened 2025-05-13
feat: Add Azure Repos as credentials provider

### LINKED ISSUE #512: Credentials add missing azurerepos option
The API supports adding the azurerepos credential type, but this is missing from the CLI. 

The fields are `username` and `password`,  in addition to `name` and `base-url`. The relevant documentation can be found here: https://docs.seqera.io/platform-cloud/git/overview#azure-devops-repositories

adamrtalbot:
Azure Repos was missing as a credentials provider despite being present in the API spec.

I can't test locally because of the private SDK so this is an educated guess, but worth a try 🤷 

fixes #512


jordeu:
@adamrtalbot is this still worth it to move it forward? I can check it.

Issue #517, opened 2025-05-29
Error exporting seqera-comute env with 0.11.5


FriederikeHanssen:
I am unable to export the seqera compute environment with:

```
tw compute-envs export -w "$ORGANIZATION_NAME/$WORKSPACE_NAME" -n seqera-compute     


Exception in thread "main" org.graalvm.nativeimage.MissingReflectionRegistrationError: The program tried to reflectively invoke method public java.lang.String io.seqera.tower.model.SeqeraComputeConfig.getDiscriminator() without it being registered for runtime reflection. Add public java.lang.String io.seqera.tower.model.SeqeraComputeConfig.getDiscriminator() to the reflection metadata to solve this problem. See https://www.graalvm.org/latest/reference-manual/native-image/metadata/#reflection for help.
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.reflect.MissingReflectionRegistrationUtils.forQueriedOnlyExecutable(MissingReflectionRegistrationUtils.java:83)
        at java.base@21.0.7/java.lang.reflect.Method.acquireMethodAccessor(Method.java:77)
        at java.base@21.0.7/java.lang.reflect.Method.invoke(Method.java:577)
        at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:688)
        at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:722)
        at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeWithType(BeanSerializerBase.java:607)
        at com.fasterxml.jackson.databind.ser.impl.TypeWrappedSerializer.serialize(TypeWrappedSerializer.java:32)
        at com.fasterxml.jackson.databind.ser.DefaultSerializerProvider._serialize(DefaultSerializerProvider.java:480)
        at com.fasterxml.jackson.databind.ser.DefaultSerializerProvider.serializeValue(DefaultSerializerProvider.java:319)
        at com.fasterxml.jackson.databind.ObjectWriter$Prefetch.serialize(ObjectWriter.java:1433)
        at com.fasterxml.jackson.databind.ObjectWriter._configAndWriteValue(ObjectWriter.java:1135)
        at com.fasterxml.jackson.databind.ObjectWriter.writeValueAsString(ObjectWriter.java:1005)
[...]

PR #518, opened 2025-05-29
bugfix(PLAT-2731): added all AwsBatchConfig parameters and some other ones until it worked

ramonamela:
solves https://github.com/seqeralabs/tower-cli/issues/517


georgi-seqera (approved):
Looks good!

This type of bug where reflection config is missing has been frequently occurring lately and I've been looking to add a step in the build to make the reflect-config generation more extensive and systematic.

tcrespog (approved):

PR #525, opened 2025-06-24
feat(PLAT-2723): Support creation of Seqera Compute environments

tom-seqera:
Support creation of Seqera Compute environment using tower-cli.

Note: because Seqera Compute automatically creates S3 buckets for CEs, `--work-dir` is optional and has a slightly different meaning. So the `CommandLine.Option` field had to be moved out of `AbstractPlatform` and instead added to each platform individually.

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/computeenvs/platforms/AltairPlatform.java
+++ b/src/main/java/io/seqera/tower/cli/commands/computeenvs/platforms/AltairPlatform.java
@@ -27,6 +27,9 @@ import java.io.IOException;
 
 public class AltairPlatform extends AbstractPlatform<AltairPbsComputeConfig> {
 
+    @Option(names = {"--work-dir"}, description = "Work directory.", required = true)
+    public String workDir;
+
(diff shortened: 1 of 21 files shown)


ramonamela (approved):

src/main/java/io/seqera/tower/cli/commands/computeenvs/platforms/AltairPlatform.java
```diff
@@ -27,6 +27,9 @@
 
 public class AltairPlatform extends AbstractPlatform<AltairPbsComputeConfig> {
 
+    @Option(names = {"--work-dir"}, description = "Work directory.", required = true)
```
cristianrcv, line 30: Would it be worth adding an intermediate abstract class that defines this --work-dir for all the classes except SeqeraCompute? The inheritance has one more layer but we avoid repeating this piece of code every time...

tom-seqera: I did think about that, but I was already finding the existing layers of inheritence and generic types a bit confusing to navigate so didn't want to add another one!

I don't feel strongly about it though.

cristianrcv (approved):

PR #531, opened 2025-07-10
Added response body when throwing data-link upload error

munishchouhan:
Added response body when throwing data-link upload error, because S3 returns the error message in the body
https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html


robsyme (approved):
Definite improvement, thanks Munish. 

We may need to include strategies for "HTTP/1.1 header parser received no bytes" and instances where we construct invalid request that prompt "Failed to upload file: HTTP 400" type errors, but more verbose error messages is always appreciated.

munishchouhan:
> Definite improvement, thanks Munish.
> 
> We may need to include strategies for "HTTP/1.1 header parser received no bytes" and instances where we construct invalid request that prompt "Failed to upload file: HTTP 400" type errors, but more verbose error messages is always appreciated.

for this, I think data team needs to add a retry mechanism cc @georgi-seqera

PR #534, opened 2025-08-07
Upgrade vulnerable dependencies

tcrespog:
# Description

- Bump versions of vulnerable transitive dependencies to non-vulnerable versions.
- Remove obsolete instruction from `README.md`.

## Guildeines for testing

- [ ] Build the executable (JAR and native).
- [ ] Run commands: there are no regressions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,12 +139,6 @@ tw CLI is a platform binary executable created by a native compilation from Java
 
     This ensures that SDKMan uses the tower-cli project-specific `.sdkmanrc` configuration.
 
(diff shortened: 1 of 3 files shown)


sabulous (approved):
Thanks Tomás, tested running a command with JAR and native executable and both worked fine.

daria-seqera (commented):
I see that it was already merged while I was trying to test it :smile:  
So just to mention that I've also compiled it locally and run a few commands with the native version, and it worked fine.

PR #535, opened 2025-08-07
Add permissions to GHA actions

tcrespog:
# Description

- Add permission to GHA actions that did not have them (considered a vulnerability).

## Guidelines for testing

- [ ] The GHA actions work correctly.

### DIFF
--- a/.github/workflows/lint-markdown.yml
+++ b/.github/workflows/lint-markdown.yml
@@ -3,6 +3,9 @@ on: [push, pull_request]
 
 jobs:
   Markdown:
+    permissions:
+      contents: write
+      actions: read
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


daria-seqera (approved):
There are some other workflows that don't have explicit permissions either (`test-backend`, `frontend-pr-checks`, some others), so why only add it to these two?

tcrespog:
> There are some other workflows that don't have explicit permissions either (`test-backend`, `frontend-pr-checks`, some others), so why only add it to these two?

A global config seems to apply for those: https://github.com/seqeralabs/tower-cli/blob/c8ef87ad4b402b15533fe990a896cf5d787ead5e/.github/workflows/build.yml#L12-L15

sabulous (approved):

PR #549, opened 2025-11-11
Add label notifier GH action

robnewman:
Workflow to send notifications to recipients when a label is applied to an issue.


.github/workflows/label-notifier.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  notify:
+    runs-on: tower-cli
+    steps:
+        - uses: jenschelkopf/issue-label-notification-action@1.3
+          with:
+            recipients:
+              squad-core=@andrew-dawson-seqera @tcrespog
+              squad-pipelines=atsyganov @swingingsimian
```
swampie: missing a `@` here : can I also be added to all squads? `@swampie`

swampie (changes requested):

.github/workflows/label-notifier.yml
```diff
@@ -0,0 +1,20 @@
...
+  issues:
+      types: [labeled]
+
+jobs:
+  notify:
+    runs-on: tower-cli
+    steps:
+        - uses: jenschelkopf/issue-label-notification-action@1.3
+          with:
+            recipients:
```
alvaromartmart: nit: IMO we should use teams instead of specific user handles. It will make IT happier

swampie (approved):

PR #555, opened 2025-11-25
Upgrade SDK version

tcrespog:
- Adapt CLI to latest API version.
- Fix `javax` annotations.
- Adapt to fluent API format for `ComputeConfig`.
- Accept case-insensitive enums.
- Fixes to tests: ignore order of JSON fields for some cases, refine some JSON mocks.


src/main/java/io/seqera/tower/cli/commands/computeenvs/platforms/AltairPlatform.java
```diff
@@ -79,6 +72,13 @@ public AltairPbsComputeConfig computeConfig() throws ApiException, IOException {
                 .maxQueueSize(adv().maxQueueSize)
                 .headJobOptions(adv().headJobOptions);
 
+        // Common
+        config.workDir(workDir)
+                .preRunScript(preRunScriptString())
+                .postRunScript(postRunScriptString())
+                .nextflowConfig(nextflowConfigString())
+                .environment(environmentVariables());
+
```
sabulous: is this change needed?

sabulous: Would you mind explaining the reason for this refactor?

If we need to do it, maybe we can also split the configuration of fields marked as `Main`, `Advanced` etc.

If we don't; then it's up to you, as we have more of the same pattern in other `computeConfig()` implementations.

src/main/java/io/seqera/tower/cli/commands/enums/WspRole.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2021-2023, Seqera.
```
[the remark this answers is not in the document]
tcrespog: All files contain that copyright 😬. Maybe we can update them all at once in a separate PR (making sure it's okay to do so).

sabulous (approved):
Ran a pipeline using the new sdk and it works. There are some configurations that I'm not that familiar with but in general it looks good to me.

PR #561, opened 2025-12-16
Fix missing compute env status `DELETING`

tcrespog:
## Description

- Apply patched version of the SDK that includes the recent CE status `DELETING`.

## Guidelines for testing

- [ ] Create a CE and transition it to `DELETING` status.
- [ ] Run `tw compute_envs` commands such as `tw compute_envs list` or `tw compute_envs view`: the execution is successful.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ dependencies {
     implementation 'ch.qos.logback:logback-core:1.5.19'
     implementation 'ch.qos.logback:logback-classic:1.5.19'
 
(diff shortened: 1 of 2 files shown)


georgi-seqera (approved):
Tested and verified that it works as expected.

minor nit: You could optionally update the [FormatHelper.java#L238-L255](https://github.com/seqeralabs/tower-cli/blob/master/src/main/java/io/seqera/tower/cli/utils/FormatHelper.java#L238-L255) to add some colour to the status :)
```
case DELETING:
      return ansi("@|fg(214) DELETING|@");
```

PR #565, opened 2026-01-07
Handle 403 responses with attached message

tcrespog:
## Description

- Show message attached to 403 responses if present.
- Replace previous default message with "Forbidden".

## Guidelines for testing

- [ ] Run a command as a user without required permissions (for example, `tw launch <pipeline>` as a `wsp:view` user): the expected error message is printed in the command output.


sabulous (approved):
lgtm

daria-seqera (approved):

PR #590, opened 2026-03-12
fix: download content from datalinks with nested folders [PLAT-4489]

weronikasosnowskaseqera:
# Description

https://seqera.atlassian.net/browse/PLAT-4489

`browse-tree` returns the full recursive tree of a given folder, including empty folders, ex:                                                                    
- data-explorer-integration-test/some-empty-folder/ ← folder, should be ignored
- data-explorer-integration-test/first-lvl-nested/file.txt ← file, should be downloaded                                                                                                                                                
                                                        
Folder entries (paths ending with `/`) are not downloadable and were causing a "File not found" error. Fix skips them during iteration.

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/data/links/DownloadCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/data/links/DownloadCmd.java
@@ -88,8 +88,9 @@ public class DownloadCmd extends AbstractDataLinksCmd {
                 pathInfo.add(new DataLinkFileTransferResult.SimplePathInfo(DataLinkItemType.FILE, path, 1));
             }
             else {
-                // Download each file for that prefix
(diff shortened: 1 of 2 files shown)


georgi-seqera (approved):
actually, I looked at the originally raised issue again and there seems to be an edge case around empty folders in the Platform API that does return an item representing a "folder" - if there is basically an empty folder without anything in it, so it is a "leaf" in the tree. 

```
{
     "items": [
             {
                 "path": "directory/filename1.txt",
                 "size": 106421
             },
             {
                 "path": "directory/subdirectory/filename2.txt",
                 "size": 106421
             },
             {
                 "path": "directory/empty-dir/",
                 "size": null
             }
     ]
 }
```

So the exact root cause is slightly different, but the fix is effectively the same. 
Could you add a test for this, please?

PR #593, opened 2026-03-30
feat: add --wait flag to compute-envs delete

robsyme:
## Summary

- Adds a `--wait` boolean flag to `tw compute-envs delete`
- After the DELETE API call returns, polls `describeComputeEnv()` until the CE is fully removed (404) or enters `ERRORED` state
- Backoff: 2s initial, +1s per poll, max 120s (matches existing `waitStatus` pattern)
- Prints progress dots (suppressed in JSON output mode)
- Prints error message if CE enters ERRORED state

## Context

Forge CE deletion is async since Dec 2025 (commit `be99dcf9b9`). The CE stays in `DELETING` status for ~2 minutes while AWS resources are cleaned up. Tools like seqerakit that delete-then-recreate CEs (`on_exists: overwrite`) need to wait for deletion to complete before creating the replacement.

Related: FD-7252, platform#10680

## Test plan

- [ ] `testDeleteWaitHappyPath`: DELETE 204 -> poll DELETING -> poll 404 -> exit 0
- [ ] `testDeleteWaitErrored`: DELETE 204 -> poll ERRORED -> exit 1
- [ ] Existing delete tests unchanged (`testDelete`, `testDeleteInvalidAuth`, `testDeleteNotFound`)

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/computeenvs/DeleteCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/computeenvs/DeleteCmd.java
@@ -96,9 +96,7 @@ public class DeleteCmd extends AbstractComputeEnvCmd {
                 }
 
                 if (status == ComputeEnvStatus.ERRORED) {
-                    if (showProgress) {
-                        app().getOut().println();
-                    }
+                    app().getErr().println("ERROR: Compute environment disposal failed (status: ERRORED). AWS resources may not have been cleaned up.");
                     return CommandLine.ExitCode.SOFTWARE;
                 }
 
--- a/src/main/java/io/seqera/tower/cli/commands/computeenvs/DeleteCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/computeenvs/DeleteCmd.java
@@ -30,6 +30,8 @@ import picocli.CommandLine.Command;
 import java.util.Collections;
 import java.util.concurrent.TimeUnit;
 
+import static io.seqera.tower.cli.utils.ResponseHelper.waitStatus;
+
 @Command(
         name = "delete",
         description = "Delete a compute environment."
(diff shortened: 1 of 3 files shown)


cristianrcv:
@robsyme To my knowledge, the Tower CLI functions primarily as a direct interface to our API, so introducing additional logic or complexity at the CLI level may not be the most appropriate approach. If an external tool requires waiting for a delete operation to complete, would it not be more suitable to implement that functionality within the tool itself?

stefanoboriero:
While you might have a point there Cristian, this behaviour pattern is already present in tower-cli (for example in LaunchCmd https://github.com/seqeralabs/tower-cli/blob/master/src/main/java/io/seqera/tower/cli/commands/LaunchCmd.java#L90-L91) so I think having it for CE deletion as well fits the current interaction pattern between seqerakit and tower-cli

src/main/java/io/seqera/tower/cli/commands/computeenvs/DeleteCmd.java
```diff
@@ -60,4 +73,52 @@ protected Response exec() throws ApiException {
...
+                }
+
+                TimeUnit.MILLISECONDS.sleep(sleepMillis);
+                sleepMillis = Math.min(sleepMillis + 1000, 120_000);
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return exitCode;
+        }
+    }
```
stefanoboriero, line 96: I'd try to reuse this existing helper function if possible https://github.com/seqeralabs/tower-cli/blob/master/src/main/java/io/seqera/tower/cli/utils/ResponseHelper.java#L137

stefanoboriero: Done in https://github.com/seqeralabs/tower-cli/pull/593/changes/ce11e285f95c5e63e0af7b47389c4fd0185964b5

stefanoboriero (approved):

PR #629, opened 2026-05-21
Soften API version mismatch from error to warning in tw info

robsyme:
## Summary

Originating from FD-7550: the `tw info` API-version check currently renders as a red `FAILED` with the message *"Tower API version is X while the minimum required version to be fully compatible is Y"*, and exits non-zero. Customers read this as a blocking incompatibility, even though the CLI still works for most operations against the older Platform.

This PR reframes the check as a compatibility warning:

- Health table shows yellow `WARNING` instead of red `FAILED` for the version row.
- New message states the mismatch, the practical impact, and the two remediations (upgrade Platform, or install an older `tw` matching the Platform version).
- `getExitCode()` no longer fails on a version mismatch — only connection or auth failures produce a non-zero exit. Scripts running `tw info` as a health check no longer break for a non-fatal condition.

Example output:

```
Tower API version check                              WARNING

    Warning: Platform API version 1.20 is older than this CLI's minimum (1.30).
    Some tw commands may fail or return incomplete results.
    To ensure full compatibility, upgrade Seqera Platform, or install a tw
    release matching your Platform version:
    https://github.com/seqeralabs/tower-cli/releases
```

## Test plan

- [x] `./gradlew test` passes (existing `InfoCmdTest` does not assert on the warning text)
- [ ] Manually run `tw info` against a Platform older than the CLI's `versionApi` and confirm yellow WARNING + new message
- [x] Manually run `tw info` against a compatible Platform and confirm green OK
- [ ] Confirm `tw info` exit code is 0 when only the version check warns, non-zero on connection/auth failure

### DIFF
--- a/src/main/java/io/seqera/tower/cli/responses/InfoResponse.java
+++ b/src/main/java/io/seqera/tower/cli/responses/InfoResponse.java
@@ -95,7 +95,6 @@ public class InfoResponse extends Response {
 
     @Override
     public int getExitCode() {
-        // versionCheck == 0 is a warning, not a failure — don't fail the command for it
-        return (connectionCheck != 0 && credentialsCheck != 0) ? CommandLine.ExitCode.OK : CommandLine.ExitCode.SOFTWARE;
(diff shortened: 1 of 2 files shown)


stefanoboriero (approved):
LGTM

jordeu (approved):

PR #643, opened 2026-07-10
feat: consume Platform SDK 1.181.0 and surface new compute-env and credentials options

cristianrcv:
## Summary

Bumps the Tower Java SDK from `1.175.0` to `1.181.0` and surfaces the new API surface added across that range in the CLI.

Each change is an isolated commit:

- **`build(deps)`** — bump `tower-java-sdk` 1.175.0 → 1.181.0, plus a small version-catalog cleanup (drop unused `jacksonDatabindNullable`/`licenserVersion` entries and refresh the now-stale licenser comments).
- **COMP-1995** — `azure-cloud` compute environment gains `--subnets` (BYO subnet names within the vnet), mirroring the existing `aws-cloud --subnet-ids`.
- **NOTASK (API 1.177.0)** — Intelligent Compute reached Azure/Google Cloud, so `azure-cloud` and `google-cloud` now expose the same scheduler options `aws-cloud` already had: `--sched-enabled`, `--provisioning-model`, `--sched-machine-types`.
- **COMP-1809** — `--fusion-metrics` (negatable) on the eight Fusion-capable platforms only (AWS/Azure Batch forge+manual, AWS/Azure/Google Cloud). Set on the `ComputeEnvComputeConfig` wrapper; the backend only accepts it when Fusion is enabled, so it is deliberately absent from platforms that never run Fusion.
- **COMP-2002 (API 1.179.0)** — new `tw credentials validate` command wrapping `POST /credentials/{id}/validate`, with `--force`. Exits non-zero on an authoritative `INVALID` result for scripting; warns on transient provider errors.
- **fix(compute)** — resolve a pre-existing NPE in `aws-cloud` when a scheduler option is passed without `--sched-enabled` (null-safe boolean), now consistent across all three cloud platforms.

## Note on scope

This PR spans four tickets (COMP-1995, COMP-1809, COMP-2002, and a NOTASK API bump) because they are all enabled by the single SDK bump and are individually small. Kept as one commit per concern for easy review.

## Test plan

- Full suite green: **644 tests, 0 failures**.
- New/updated tests: azure-cloud `--subnets`; azure-cloud & google-cloud scheduler options; aws-cloud `--fusion-metrics` / `--no-fusion-metrics` at the wrapper level; aws-cloud scheduler-without-enable NPE regression; `credentials validate` for AVAILABLE (exit 0) and INVALID (exit 1).
- Drove the built CLI (`--help`) to confirm the new command and flags render correctly.


stefanoboriero (commented):
VERSION-API file needs to be bumped to 1.181.0, because Azure networking changes were shipped in that version

stefanoboriero (approved):
I think this also needs a little bump to make tests pass when VERSION-API is bumped https://github.com/seqeralabs/tower-cli/blob/master/src/test/resources/runcmd/info/service-info.json#L4.

Other than that looking good

PR #646, opened 2026-07-20
fix(deps): upgrade jackson-bom to 2.18.8 (COMP-1963)

cristianrcv:
## Summary
- Upgrades the Jackson BOM (`com.fasterxml.jackson:jackson-bom`) from `2.18.6` to `2.18.8`, which pins `jackson-databind` to the patched `2.18.8` release.
- Resolves a PolymorphicTypeValidator (PTV) bypass via generic type parameters that allows arbitrary class instantiation.
- Resolves CVE-2026-54512 / GHSA-j3rv-43j4-c7qm

The BOM is the directly declared dependency that manages the jackson-databind version, so bumping it is the correct fix — no transitive pinning / constraints / force were used.

## JIRA
COMP-1963: Fix jackson-databind has a PolymorphicTypeValidator bypass via generic type parameters that allows arbitrary class instantiation

## Security Advisory
https://github.com/seqeralabs/tower-cli/security/dependabot/20

🤖 Generated with [Claude Code](https://claude.ai/code)


cristianrcv:
Closed in favour of https://github.com/seqeralabs/tower-cli/pull/647

PR #653, opened 2026-07-31
feat: [COMP-2106] support VPC network and subnetworks for google-cloud compute environment

jonmarti:
## Summary
Adds VPC/subnet customization to the `compute-envs add google-cloud` command, which previously had no networking options. Mirrors the existing `aws-cloud` VPC/subnet support (#639).

New advanced options:
- `--network` — VPC network name or URI (defaults to the project's `default` network)
- `--subnetworks` — comma-separated subnetworks; the first is used for basic placement, Intelligent Compute may use all of them
- `--network-tags` — firewall-targeting tags (validated: require `--network`, GCP tag format, max 64)
- `--use-private-address` — launch VMs without an external IP

`subnetworks` is a list (not singular like `google-batch`), matching the `GoogleCloudConfig` SIC model where workers can be spread across a subnet pool.

Also bumps the minimum required Platform API version to `1.190.0` (the version that introduced the `GoogleCloudConfig` network fields, and the current cloud production API version), updating the `info` command tests/fixture to match.

### Unit tests
New tests in `GoogleCloudPlatformTest`:
- `--network` + `--subnetworks` (list) + `--use-private-address` → asserts exact request body
- `--network-tags` → asserts tags in request body
- `--network-tags` without `--network` → fails CLI-side ("Network tags require VPC configuration")
- invalid tag format (`Allow-SSH`) → fails CLI-side

### Manual end-to-end verification
Built the CLI shadow jar and exercised the real `compute-envs add google-cloud` flow against live backends, confirming each field round-trips CLI flag → API → persisted CE config.

| Scenario | Backend | Result |
|----------|---------|--------|
| network + subnetworks + tags + private-address (no SIC) | local (API 1.195.0) | CE created; persisted `network=sic-test-vpc`, `subnetworks=[sic-test-subnet]`, `networkTags=[allow-ssh, web-tier]`, `usePrivateAddress=true` ✓ |
| same, non-SIC | dev (API 1.195.0), personal workspace | same four fields persisted ✓ |
| **SIC enabled** + network flags | dev (API 1.195.0), org workspace | `schedEnabled=true`, `schedConfig={provisioningModel: ondemand, machineTypes: [n2d-standard-2]}` **and** all four network fields persisted ✓ |

Negative validation (tags without network, invalid tag format) confirmed to fail CLI-side against the live backend as well.

The SIC + VPC/subnet combination is the key case for COMP-2106 (VPC/subnet customization for GCP Cloud Intelligent Compute); verified the two work together.

Requires `tower-java-sdk` >= 1.190.0 (already pinned).

### DIFF
--- a/VERSION-API
+++ b/VERSION-API
@@ -1,4 +1,4 @@
-1.190.0
+1.189.0
 // Only first line of this file is read
 // This version should be bumped to the minimum version where dependent API changes were introduced
 // But never higher then the current Platform API Version deployed in Cloud Production: https://cloud.seqera.io/api/service-info
--- a/VERSION-API
+++ b/VERSION-API
@@ -1,4 +1,4 @@
-1.181.0
+1.190.0
 // Only first line of this file is read
 // This version should be bumped to the minimum version where dependent API changes were introduced
 // But never higher then the current Platform API Version deployed in Cloud Production: https://cloud.seqera.io/api/service-info
--- a/src/main/java/io/seqera/tower/cli/commands/computeenvs/platforms/GoogleCloudPlatform.java
+++ b/src/main/java/io/seqera/tower/cli/commands/computeenvs/platforms/GoogleCloudPlatform.java
@@ -17,6 +17,7 @@
 package io.seqera.tower.cli.commands.computeenvs.platforms;
 
 import io.seqera.tower.ApiException;
+import io.seqera.tower.cli.exceptions.TowerRuntimeException;
 import io.seqera.tower.model.ComputeEnvComputeConfig.PlatformEnum;
 import io.seqera.tower.model.GoogleCloudConfig;
 import io.seqera.tower.model.SchedConfig;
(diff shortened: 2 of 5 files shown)


stefanoboriero:
Remember to update the minimal required API version for this to work in https://github.com/seqeralabs/tower-cli/blob/master/VERSION-API

jonmarti:
> Remember to update the minimal required API version for this to work in https://github.com/seqeralabs/tower-cli/blob/master/VERSION-API

@stefanoboriero thanks for the heads-up , addressed in e8bda08 9647133. Bumped to 1.190.0, the version that introduced the `GoogleCloudConfig` network fields and the current cloud production API version

ramonamela (approved):

VERSION-API
```diff
@@ -1,4 +1,4 @@
-1.181.0
+1.190.0
```
stefanoboriero: Are you sure it's not 1.189.0? https://github.com/seqeralabs/platform/blob/master/changelog-API.md#11890

stefanoboriero (approved):

VERSION-API
jonmarti: Good catch Stefano, ty

PR #649, opened 2026-07-22
feat: [PLAT-5936] Allow-list support for private studios

sabulous:
###  Description

Ticket: [PLAT-5936](https://seqera.atlassian.net/browse/PLAT-5936)

This PR adds CLI support for the private-studio **allow list** — the users, besides the creator, who may connect to and start a private studio.

### Changes

- **Set the list on add** — Set the allowed user when **creating** a private studio with the `--allow-user <user>` option (available on add, add-as-new, and start), where `<user>` is a **numeric user ID, username, or email**. The platform currently caps the allow list at a single additional user, so the option takes a single value.
   ```bash
   tw studios add -n my-studio -w my-workspace -c my-ce -t <template> --private --allow-user <user>
   ```
- **Name/email resolution** — a non-numeric `--allow-user` value is resolved to a user ID by looking it up among the organization's members and then its collaborators (a workspace participant may be either). A numeric value is used directly. The backend still enforces that the resolved user is a workspace participant.

- **Override the list on start** — `start --allow-user` **replaces** the existing allow list; omit it to leave the list unchanged. Only the studio creator may change it.
_Note: the override only applies when the studio is **startable** (stopped) — starting an already-running studio won't change the list._
  ```bash
  tw studios start -w my-workspace -n my-studio --allow-user <user>
  ```

- **Studio view shows the allow list** — the studio view now includes an **Allowed users** row.

- **Lookup-by-name returns full detail** — `view -n` (and `start`/`update` by name) now fetch the full studio via `describe` after resolving the name, so name-based lookups match id-based ones (the list response omits `allowedUsers`).
  ```bash
  tw studios view -w my-workspace -n my-studio
  ```

<details>

<summary>Example cli testing session with various cases</summary>

> **Setup:** These are real runs against a local backend, workspace `hello1 / 1-wsp`. The studio `cli-priv-1` is created by user `sab`. `21` is user `nonowner` and `8` is user `permiss`, both workspace members. `21` is added to the allow list at creation; `8` replaces it later via `start`.

### Create a private studio with an allowed user (by id)
```console
$ tw studios add -w 152937999146850 -n cli-priv-1 -t cr.seqera.io/public/data-studio-jupyter:4.2.5-snapshot -c nwelocal --private --allow-user 21

  Studio 9def339e CREATED at [hello1 / 1-wsp] workspace.
```

### The allowed user can also be given by username or email
```console
[...]

PR #656, opened 2026-08-04
chore: Bump VERSION-API to 1.190.0

sabulous:
Bump `VERSION-API` to `1.190.0` that supports allow list for studios.

### DIFF
--- a/VERSION-API
+++ b/VERSION-API
@@ -1,4 +1,4 @@
-1.189.0
+1.190.0
 // Only first line of this file is read
 // This version should be bumped to the minimum version where dependent API changes were introduced
 // But never higher then the current Platform API Version deployed in Cloud Production: https://cloud.seqera.io/api/service-info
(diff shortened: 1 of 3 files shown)


stefanoboriero:
Judging from platform api changelog, this was introduced in api version 1.186.0 https://github.com/seqeralabs/platform/blame/master/changelog-API.md#L335 so leaving 1.189.0 is enough (it was recently bumped in https://github.com/seqeralabs/tower-cli/pull/653). This value is documenting the minimum api version a seqera installation should have in order to be fully compatible with the given cli version. It is not intended to match the tower-sdk dependency version, which could be higher than the minimal API version requirement if features are not used.

My recommendation is to close this PR, ideally the change https://github.com/seqeralabs/tower-cli/pull/649 should have increased from 1.181.0 to 1.186.0 but the subsequent bump from the GPC networking options got us covered

PR #671, opened 2026-08-25
fix(launch): align CLI launch requests with the API model [COMP-2318]

cristianrcv:
Fixes [COMP-2318](https://seqera.atlassian.net/browse/COMP-2318) (support ticket FD-7868).

Three commits: the silent-drop bug, then `--syntax-parser`, then `--nextflow-version` and `--output-dir`. The ticket filed the options as separate follow-ups, but the customer's actual workflow — registering pipelines across several workspaces with `tw pipelines add` — cannot select v2 without them, and the other two fields had the same gap.

## The bug

`WorkflowLaunchRequest` objects were built from **five** independent hand-maintained field lists, all drifted from the generated SDK model and all omitting `syntaxParser`. Platform coerces an omitted `syntaxParser` to `v1` (companion bug PLAT-6585), so dropping the field is equivalent to explicitly requesting the legacy parser: a pipeline set to v2 in the web UI ran under v1 from the CLI, with no warning anywhere.

The ticket named two mappers. There were five:

| Mapper | Used by | Was |
| --- | --- | --- |
| `ModelHelper.createLaunchRequest` | `tw launch`, `pipelines export`, `pipelines view`, `actions view` | 27 of 31 fields |
| `LaunchCmd.updateLaunchRequest` | `tw launch` (both forms) | **not in the ticket** — rebuilt the request right after `ModelHelper` had built it |
| `RelaunchCmd.exec` | `tw runs relaunch` | own list |
| `pipelines UpdateCmd.buildUpdateRequest` | `tw pipelines update` | **not in the ticket** — any single-field update reset everything omitted |
| `actions UpdateCmd.exec` | `tw actions update` | **not in the ticket** — same |

The `LaunchCmd` one mattered most: `runTowerPipeline` pipes `createLaunchRequest`'s output straight through `updateLaunchRequest`, so fixing `ModelHelper` alone would have left `tw launch` broken. The two update commands mean a v2 pipeline was also downgraded by an unrelated `tw pipelines update`.

## The fix

**`ModelHelper.createLaunchRequest`** copies the configuration by serializing the source model and reading it back as a request, instead of naming fields. Every field the models share is carried over, and fields the API gains in a future SDK release need no change here — which is the point, since these lists had already drifted independently five times over.

Going through JSON is also what makes the nullable fields correct. `syntaxParser`, `nextflowVersion` and `outputDir` are `JsonNullable`, and the fluent setters wrap their argument unconditionally:

\`\`\`java
public WorkflowLaunchRequest syntaxParser(SyntaxParserEnum syntaxParser) {
  this.syntaxParser = JsonNullable.<SyntaxParserEnum>of(syntaxParser);   // null becomes an explicit null
\`\`\`

So the obvious one-line fix (`.syntaxParser(launch.getSyntaxParser())`) would have started sending `"syntaxParser": null` for every pipeline that never set it — which Platform coerces to `v1` just the same, i.e. the same bug wearing a different hat. The round-trip keeps *undefined* undefined; `ModelHelperTest` pins that in both directions. The same trap is why the new options are applied only when passed, via `LaunchOptions.applyNullableOptions`, which documents the reason once instead of in six commands.

**The four command-level mappers** now start from the stored configuration and apply only the overrides the user asked for, so each names just the fields its own flags can change.

## The options

`--syntax-parser v1|v2`, `--nextflow-version` and `--output-dir`, on `tw pipelines add|update`, `tw actions add|update`, `tw launch` and `tw runs relaunch` — in `LaunchOptions` for the six sharing that mixin, plus `LaunchCmd`, which declares its own. Verified in the built jar: all three appear in all six commands' help.

All three are writable in the API spec (`seqera-api.yml`) — no `readOnly`, which the spec does use 39 times elsewhere, and `CreatePipelineRequest`, `UpdatePipelineRequest`, `CreateActionRequest`, `UpdateActionRequest` and `SubmitWorkflowLaunchRequest` all `$ref` the same `WorkflowLaunchRequest`. Per the spec, `nextflowVersion` must exist in the version catalog and meet the compute environment's minimum, and `outputDir` needs Nextflow ≥ 24.10 with the workflow outputs syntax; both are validated server-side, so a bad value surfaces as an API error.

No type converter for the enum: `Tower.buildCommandLine` already calls `setCaseInsensitiveEnumValuesAllowed(true)`, so `v2` resolves to `SyntaxParserEnum.V2` and picocli rejects anything else while listing the accepted values (`testAddWithInvalidSyntaxParser` pins that). I wrote a converter first and deleted it — worth knowing the setting is there before adding one.

## Fields recovered beyond the options

`launchContainer`, `pipelineSchemaId`, and on the update paths everything the partial lists omitted. (`lineage`, the fourth name in the ticket, does not exist in any SDK 1.200.0 launch model — nothing to fix.)

## Behaviour changes a reviewer should weigh

All deliberate; say the word and I will narrow any of them.

- **`tw pipelines update` and `tw actions update` no longer clear stored fields.** In `actions update` that stops `--revision`, `--profile`, `--params-file`, `--config`, `--pull-latest`, `--stub-run`, `--main-script` and `--entry-name` from resetting their stored values when omitted. This is the largest change in the PR and it is a pre-existing bug of the same family, found while adding the options to those commands.
- **`tw launch` keeps `--launch-container` unset instead of clearing the stored value**, which it wiped whenever the flag was absent.
- **`tw runs relaunch` now inherits optimization settings and head job resources** from the original run. This follows from "relaunch the same configuration", but `relaunch` has no `--disable-optimization` to opt out the way `launch` does. Easy to drop if you would rather leave relaunch alone.
- **`tw runs relaunch` explicitly clears `commitId`.** `revision` already carries the commit to run (the resume commit when resuming), so an inherited `commitId` would pin the relaunch to the original commit and silently defeat both resume and `--revision`. Matches today's payload.
- **`tw pipelines export` output gains keys** — that is the fix: re-importing an exported v2 pipeline no longer downgrades it. `pipelines view` and `actions view` serialize the same request and now show these fields too.
- **`tw launch` still clears `sessionId`, `resume` and `dateCreated`** — a new run starts a fresh session and never resumes — so its payload is otherwise unchanged.

## How to verify

\`\`\`
./gradlew test
\`\`\`

698 tests, 0 failures. Note the build needs GitHub Packages credentials for `tower-java-sdk:1.200.0` (`github.packages.user` / `github.packages.token`, or `GITHUB_USERNAME` / `GITHUB_TOKEN`).

Eleven new tests. The ten command-level ones were run against `master`'s `src/main` with this branch's tests in place, and **all ten fail there** (three unrelated tests also fail in that configuration, because a command rejecting an unknown option leaves its mocks unconsumed for the next test; they pass on the branch):

- `ModelHelperTest` — full field copy, the `WorkflowLaunchResponse` overload, and undefined vs explicit-null nullable behaviour.
- `LaunchCmdTest` — the real `POST /workflow/launch` body carries the stored `syntaxParser: v2`, `sessionId`/`resume` stay absent, and all three options override.
- `RunsCmdTest` — same for relaunch, plus no inherited `commitId`, plus the override.
- `PipelinesCmdTest` — `add` with each option, an invalid enum value rejected, `update` preserving a stored v2 with no flag, `update --syntax-parser v1` overriding, and the exported JSON.

That last one matters: the pre-existing `testExport` builds its expected output by calling `ModelHelper` itself, so it could never have caught this class of bug, which is part of why the drift went unnoticed.

[COMP-2318]: https://seqera.atlassian.net/browse/COMP-2318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/seqera/tower/cli/commands/LaunchCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/LaunchCmd.java
@@ -20,6 +20,7 @@ import io.seqera.tower.ApiException;
 import io.seqera.tower.cli.commands.enums.OutputType;
 import io.seqera.tower.cli.commands.global.WorkspaceOptionalOptions;
 import io.seqera.tower.cli.commands.labels.Label;
+import io.seqera.tower.cli.commands.pipelines.LaunchOptions;
 import io.seqera.tower.cli.commands.pipelines.versions.VersionRefOptions;
 import io.seqera.tower.cli.exceptions.InvalidResponseException;
 import io.seqera.tower.cli.responses.Response;
@@ -100,6 +101,12 @@ public class LaunchCmd extends AbstractRootCmd {
     @Option(names = {"--syntax-parser"}, description = "Nextflow language syntax parser version: 'v1' (legacy) or 'v2'. Overrides the value stored in the pipeline.")
     SyntaxParserEnum syntaxParser;
 
+    @Option(names = {"--nextflow-version"}, description = "Nextflow version to run the workflow with. Must exist in the Platform version catalog and meet the minimum required by the compute environment. Overrides the value stored in the pipeline.")
+    String nextflowVersion;
+
+    @Option(names = {"--output-dir"}, description = "Per-run output directory, passed to Nextflow as '-output-dir'. Requires Nextflow 24.10.0 or later and the workflow outputs syntax. Overrides the value stored in the pipeline.")
+    String outputDir;
+
     @ArgGroup(heading = "%nAdvanced options:%n", validate = false)
     AdvancedOptions adv;
 
@@ -144,11 +151,7 @@ public class LaunchCmd extends AbstractRootCmd {
     private WorkflowLaunchRequest updateLaunchRequest(WorkflowLaunchRequest base) throws IOException {
         boolean disableOptimization = coalesce(adv().disableOptimization, false);
 
-        // Only set when requested: the setter cannot express 'undefined', so an unconditional call
-        // would replace the pipeline's stored value with an explicit null, which Platform reads as v1.
-        if (syntaxParser != null) {
-            base.syntaxParser(syntaxParser);
-        }
+        applyNullableOptions(base);
 
         return base
                 .runName(name)
@@ -176,6 +179,23 @@ public class LaunchCmd extends AbstractRootCmd {
                 .sessionId(null)
                 .resume(null)
                 .dateCreated(null);
+    }
+
+    /**
+     * Same reasoning as {@link LaunchOptions#applyNullableOptions}: these setters wrap their argument in
+     * {@code JsonNullable.of()}, so applying them unconditionally would overwrite the pipeline's stored
+     * value with an explicit null. This command declares its own options rather than the shared mixin.
+     */
+    private void applyNullableOptions(WorkflowLaunchRequest base) {
+        if (syntaxParser != null) {
+            base.syntaxParser(syntaxParser);
+        }
+        if (nextflowVersion != null) {
+            base.nextflowVersion(nextflowVersion);
+        }
+        if (outputDir != null) {
+            base.outputDir(outputDir);
+        }
     }
 
     protected Response runTowerPipeline(Long wspId) throws ApiException, IOException {
--- a/src/main/java/io/seqera/tower/cli/commands/LaunchCmd.java
+++ b/src/main/java/io/seqera/tower/cli/commands/LaunchCmd.java
@@ -31,6 +31,7 @@ import io.seqera.tower.model.PipelineDbDto;
 import io.seqera.tower.model.SubmitWorkflowLaunchRequest;
 import io.seqera.tower.model.SubmitWorkflowLaunchResponse;
 import io.seqera.tower.model.WorkflowLaunchRequest;
+import io.seqera.tower.model.WorkflowLaunchRequest.SyntaxParserEnum;
 import io.seqera.tower.model.WorkflowStatus;
 import jakarta.annotation.Nullable;
 import picocli.CommandLine;
@@ -96,6 +97,9 @@ public class LaunchCmd extends AbstractRootCmd {
     @Option(names = {"--launch-container"}, description = "Container image to use for the Nextflow launcher.")
     String launchContainer;
 
+    @Option(names = {"--syntax-parser"}, description = "Nextflow language syntax parser version: 'v1' (legacy) or 'v2'. Overrides the value stored in the pipeline.")
+    SyntaxParserEnum syntaxParser;
+
     @ArgGroup(heading = "%nAdvanced options:%n", validate = false)
     AdvancedOptions adv;
 
@@ -139,6 +143,12 @@ public class LaunchCmd extends AbstractRootCmd {
      */
     private WorkflowLaunchRequest updateLaunchRequest(WorkflowLaunchRequest base) throws IOException {
         boolean disableOptimization = coalesce(adv().disableOptimization, false);
+
+        // Only set when requested: the setter cannot express 'undefined', so an unconditional call
+        // would replace the pipeline's stored value with an explicit null, which Platform reads as v1.
+        if (syntaxParser != null) {
+            base.syntaxParser(syntaxParser);
+        }
 
         return base
                 .runName(name)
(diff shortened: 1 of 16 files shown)


src/main/java/io/seqera/tower/cli/commands/pipelines/LaunchOptions.java
```diff
@@ -70,4 +72,31 @@ public class LaunchOptions {
...
+        if (syntaxParser != null) {
+            request.syntaxParser(syntaxParser);
+        }
+        if (nextflowVersion != null) {
+            request.nextflowVersion(nextflowVersion);
+        }
+        if (outputDir != null) {
+            request.outputDir(outputDir);
+        }
+    }
```
alberto-miranda: Not blocking for this PR but this looks like a landmine waiting for poor soul to trip on it. What if we forget to call or extend this function as the code evolves? This looks like a detail that should be taken care of by `coalesce()` internally.

src/main/java/io/seqera/tower/cli/commands/LaunchCmd.java
```diff
@@ -154,12 +170,32 @@ private WorkflowLaunchRequest updateLaunchRequest(WorkflowLaunchRequest base) th
...
+        if (syntaxParser != null) {
+            base.syntaxParser(syntaxParser);
+        }
+        if (nextflowVersion != null) {
+            base.nextflowVersion(nextflowVersion);
+        }
+        if (outputDir != null) {
+            base.outputDir(outputDir);
+        }
     }
```
alberto-miranda: The code is identical to its `LaunchOptions.java` counterpart. Should we be worried about drift here? Any chance for a refactor to remove the duplication?

alberto-miranda (approved):
Left two minor comments that might be worth looking into, but nothing blocking.

src/main/java/io/seqera/tower/cli/commands/pipelines/LaunchOptions.java
cristianrcv: The problem is that coalesce cannot distinguish null vs not provided

src/main/java/io/seqera/tower/cli/commands/LaunchCmd.java
cristianrcv: That I can try :)

cristianrcv: Fixed in [4414c27](https://github.com/seqeralabs/tower-cli/pull/671/commits/4414c27c50e57ec456bb38f82b7532af6e111106)

src/main/java/io/seqera/tower/cli/commands/pipelines/LaunchOptions.java
cristianrcv: This I cannot fix but I added a test so that new variables do not skip this

src/main/java/io/seqera/tower/cli/commands/pipelines/NullableLaunchOptions.java
```diff
@@ -0,0 +1,60 @@
...
+ * (COMP-2318). {@code coalesce} cannot paper over this: by the time it runs, "the user passed nothing" and
+ * "the user passed null" are the same value. So each option is applied only when it was actually given,
+ * by {@link #applyNullableOptions}.
+ * <p>
+ * {@link LaunchOptions} extends this class so the commands mixing it in inherit these options; commands
+ * declaring their own option set mix this class in directly. Adding an option here without handling it in
+ * {@link #applyNullableOptions} would make it silently do nothing, so a test enforces that every option
+ * declared here is applied.
+ */
+public class NullableLaunchOptions {
```
stefanoboriero, line 38: These are Nullable in the sense that they are marked as nullable in the OpenApi schema, and because of that the autogenerated SDK generates the JsonNullable.of shenanigan?

stefanoboriero (approved):

src/main/java/io/seqera/tower/cli/commands/pipelines/NullableLaunchOptions.java
cristianrcv: Yes. The idea is at least to not have to copy every single new parameter into 3 different places but rather only add these that are "exceptions"

